PR #72, opened 2021-02-11
Feature: Certification Tab 

alewoo28:
**Linked Issue:**
https://qmacbis.atlassian.net/browse/OY2-3892

**Overview:**
The certification tab is a static/default tab in the tab container (along with the summary tab). In the certification tab users have the option of provisional submission and final submission. Both buttons change the status of the form and trigger a save. (Actual save functionality has yet to be implemented)

**Files Changed (6):** 
`services/ui-src/src/components/SummaryTab.js`
The summary tab is the other static tab. It has dummy data in it right now until it is further built out 

`services/ui-src/src/components/layout/TabContainer.js`
The tab container has been updated to host both the SummaryTab and CertificationTab as components instead of divs with text. 

`services/ui-src/src/components/CertificationTab.js`
This component houses both of the buttons for certification options. 

When a user selects "Certify & Submit Provisional Data" it triggers certifyAndSubmitProvisional() which updates the state of the form to provisional (and should save) and disables the button.  In redux the form status is updated as well as last_modified_by and last_modified.  An alert is displayed reflecting this. 

When a user selects "Certify & Submit Final Data" it triggers certifyAndSubmitFinal() which updates the state of the form to final (and should save) and disables both buttons.  in redux the form status is updated as well as last_modified_by and last_modified. An alert is displayed reflecting this 

Both buttons' disable feature is toggled based on the status. The provisional button should be disabled if status is provisional OR final. The final button should be disabled when the status is final. 

`services/ui-src/src/store/actions/certify.js`
This is where all of the action creators and thunk functions for certification live. 
`certifyAndSubmitFinal()` will update the store (using `setFinalCertify()`), and then trigger a save. 
`certifyAndSubmitProvisional()` will update the store (using `setProvisionalCertify()`), and then trigger a save. 

`services/ui-src/src/store/reducers/singleForm.js`
Actions were added to the reducer.  `CERTIFY_AND_SUBMIT_FINAL` & `CERTIFY_AND_SUBMIT_PROVISIONAL` do virtually the same thing. It will be up to the frontend to know that a form in status "final" is not editable. `UNCERTIFY_FORM` will update the status of the form back to 'in_progress' but we dont currently have anywhere we need to put this for right now.

`services/ui-src/src/scss/_tabs.scss` 
Added some styling to the certification tab, specifically on the buttons


alewoo28:
@jgillis1 Can you look at the wording in this tab for me? The guidance in the ticket only mentioned copying some of the words from carts, so some of it was improvised. 

I'm not sure what the 6th changed file is, it may have something to do with debugger but I will check with the team before removing the WIP label.

jgillis1 (approved):

PR #95, opened 2021-02-19
Generate form json

eolaniyan052:
Create JSON to import into DB for Form 64.EC. PDF of form attached. Should be written out in 'grid view' as shown in the screenshot from CARTS.
image


jgillis1:
Replaces #94

jgillis1 (approved):

PR #112, opened 2021-02-24
Bugfix api users

jgillis1:
Attempting Doug's recommended solution to add the auth-users table to the LamdaApiRole policy allow permissions.

Also, change the default testing state to AL (from AK) since we have data for AL in the database to test/work against.


jgillis1:
deploy check threw this error on first push:
`An error occurred: LambdaApiRole - Resource bugfix-api-users-auth-user must be in ARN format or "*". (Service: AmazonIdentityManagement; Status Code: 400; Error Code: MalformedPolicyDocument; Request ID: c84e7e33-3176-4f6a-b836-bfd09a983320; Proxy: null).`

eolaniyan052 (approved):

PR #137, opened 2021-03-02
Create 21PW form and answer. 

TGriesemer:
For ticket 5051[(https://qmacbis.atlassian.net/secure/RapidBoard.jspa?rapidView=184&modal=detail&selectedIssue=OY2-5051)
Add age range, form record, reference to json files in serverless, and add 3 records to state_forms

### DIFF
--- a/src/database/initial_data_load/21PW.json
+++ b/src/database/initial_data_load/21PW.json
@@ -1,13 +1,13 @@
 [
   {
     "question": "2021-21PW-01",
-    "form_id": "1",
+    "form_id": "6",
     "form": "21PW",
     "year": 2021,
     "type": "datagridwithtotals",
     "row_total": true,
     "column_total": true,
-    "label": "Unduplicated # Ever Enrolled",
+    "label": "What is the unduplicated number of children &&&VARIABLE&&& ever enrolled during the quarter?",
     "rows": [
       {
         "col1": "",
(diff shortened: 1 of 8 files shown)


src/database/initial_data_load/state_forms.json
```diff
@@ -286,5 +286,76 @@
     "last_modified_by": "seed",
     "created_date": "01-15-2021",
     "created_by": "seed"
+  },{
+  "state_form": "PA-2021-1-21PW",
+  "state_id": "PA",
+  "form_id": "3",
```
jgillis1, line 292: Could you correct the form_id throughout state_forms.json? It should be 6 it looks like.

src/database/initial_data_load/21PW.json
```diff
@@ -0,0 +1,580 @@
+[
+  {
+    "question": "2021-21PW-01",
+    "form_id": "1",
```
jgillis1, line 4: Can you fix the form_id here as well to match what you entered in forms?

src/database/initial_data_load/21PW.json
```diff
@@ -0,0 +1,580 @@
...
+    "form": "21PW",
+    "year": 2021,
+    "type": "datagridwithtotals",
+    "row_total": true,
+    "column_total": true,
+    "label": "Unduplicated # Ever Enrolled",
```
jgillis1, line 10: Could we use the long version of the questions for the label like we do in 21E?

src/database/initial_data_load/form_answers_21PW_AL.json
```diff
@@ -0,0 +1,543 @@
+[
+  {
+    "answer_entry": "AL-2021-1-21PW-U1-01",
```
jgillis1, line 3: Please change "U1" in the ID to be "64" since that's the high end of the age range this question applies to

jgillis1: Please do that throughout this file and the samples for the other states as well.

jgillis1 (approved):

PR #140, opened 2021-03-02
Homestate

KodyLincoln:
ticket oy2-4573


services/ui-src/src/containers/HomeState.js
```diff
@@ -32,34 +56,30 @@ const HomeState = () => {
       <ul className="quarterly-items">
         {years[year].quarters.map(element => {
           return (
-            <li key={`${state}-${year}-${element}`}>
-              <Link href={`/forms/${state}/${years[year].year}/${element}`}>
+            <li key={`${state}-${years[year].year}-${element}`}>
+              <Link href={`/forms/${state}/${years[year].year}./${element}`}>
```
jgillis1, line 60: I think the . before /${element} needs to be removed as it's generating the URL looking like this: http://localhost:3000/forms//2021./1

services/ui-src/src/containers/HomeState.js
```diff
@@ -1,29 +1,53 @@
 import React from "react";
 import { Accordion, Grid, Link } from "@trussworks/react-uswds";
+import { connect } from "react-redux";
 
-const HomeState = () => {
-  // TODO: Pull state from redux
-  const state = "AL";
+const HomeState = ({ userData }) => {
+  // Pull state from redux
+  const state = userData.abbrev;
```
jgillis1, line 7: ${state} is not being populated in the link when you click on the Quarter you want to view. Could you follow up to see why? In LocalLogin.js I changed the info for Alice Cooper to be role: "state" and states: ["AL"] only so it should have picked up Alabama and populated that when moving from this page to the Quarter page.

services/ui-src/src/containers/HomeState.js
```diff
@@ -1,29 +1,53 @@
...
-      year: 2019,
-      quarters: ["1", "2", "3", "4"]
-    },
-    {
-      year: 2018,
-      quarters: ["1", "2", "3", "4"]
+      if (i === fiscalYear) {
+        currentYear = fiscalYear;
+        if (month > 9) {
+          currentYear = fiscalYear + 1;
```
jgillis1, line 26: This is cleverly done.

jgillis1 (changes requested):

mblevins-collabralink (changes requested):
DItto to what Jenna said.

KodyLincoln:
Homestate.js is used to route "state" users to their state forms from their quarterly reports page upon logging into the system.

This ensures state users are subject to only the state role.

Homestate is enabled when Home.js reads that the role of the user is "state" and the state that the user belongs to upon log in and passes them onto their states quarterly form page. The quarters are generated by taking the date and transforming it into 4 fiscal quarters that begin with 2018 and will continue to populate as time moves forward. When the link in quarter is clicked the state is passed as a property from Home.js and the year and quarter based on the selection on the page.

Code climate has failed this twice based on the code being too long.

PR #146, opened 2021-03-03
add set stages for termination protection

eolaniyan052:
Purpose of this PR is to fix the issue with termination protection on every stack which is preventing them to delete, added a termination stages on each layer of our application


mdial89f (approved):
Looks good here

PR #161, opened 2021-03-09
Feature: Form Page (redux version)

alewoo28:
**Linked Issue:**
https://qmacbis.atlassian.net/browse/OY2-5892

**Overview:**
The Form page dynamically displays the form selected by the user on the quarter reports page. The FormPage component houses all of the components. From FormPage to question its FormPage>TabContainer>TabPanel>QuestionComponent>GridWithTotals. Each grid is stateful and houses its own state. 

There are a lot of files changed unfortunately so I've been as verbose as possible to make it easier to sort through. If anything is unclear I'm happy to hop on a quick call to clarify or make any changes. 
Fortunately half of the changes files are only JSON changes. 

**Files Changed (31):**

**_JSON CHANGES(14):_**
`src/database/initial_data_load/form_answers_64_21E_PA.json`
`src/database/initial_data_load/form_answers_64_21E_MD.json`
`src/database/initial_data_load/form_answers_64_21E_AL.json`
`src/database/initial_data_load/form_answers_64EC_PA.json`
`src/database/initial_data_load/form_answers_64EC_MD.json`
`src/database/initial_data_load/form_answers_64EC_AL.json`
`src/database/initial_data_load/form_answers_21PW_PA_Q4.json`
`src/database/initial_data_load/form_answers_21PW_PA.json`
`src/database/initial_data_load/form_answers_21PW_MD.json`
`src/database/initial_data_load/form_answers_21PW_AL.json`
`src/database/initial_data_load/form_answers_21E_PA.json`
`src/database/initial_data_load/form_answers_21E_MD.json`
`src/database/initial_data_load/form_answers_21E_AL.json`
All 13 of these are just the seed files. Tim switched to my branch and changed the IDs for me after group discussion. We changed ID format from two digits to four. ie: 
  `"answer_entry": "AL-2021-1-21E-01-01"` to `"answer_entry": "PA-2021-1-64.21E-0000-01"`
Confluence has been updated to reflect these changes. The age ranges in the ID string now also matches the rangeIDs in the age-range table. Changes to lines 289-312 are just linter changes. 

`src/database/initial_data_load/state_forms.json`
A few of the form ID's were incorrect in our seed data. Some application logic relies on there being only one of each form in a given state/year/quarter  so it was important to fix.

**_REDUX WORK (4)_**
`services/ui-src/src/store/helperFunctions.js`
I made a file for helper functions specific to redux. The frontend will render the questions in whatever order they are when they get there. `sortQuestionsByNumber` was too bulky to put inside of a Thunk so I made it its own function. 
Its meant to be used with the Array.sort method. It takes in two elements at a time, extracts their question IDs, compares the last numbers and returns the two elements sorted. 

`services/ui-src/src/store/reducers/global.js`
The getFormTypes api populates the global reducer using the fetchFormTypes thunk. Static dummy data from JSON removed.

`services/ui-src/src/store/reducers/quarterStatuses.js`
The getStateForms api populates the quarterStatuses reducer using the getQuarterStatuses thunk. Static dummy data from JSON removed.

`services/ui-src/src/store/reducers/singleForm.js`
The getSingleForm and getStateForms api populate the singleForm reducer using the getFormData thunk. Questions & answers are fetched from the getSingleForm api. Questions are sorted with a helper function. Status data is fetched from the getStateForms endpoint and the status data for the current form is filtered out. Answers, questions and statusData are dispatched to the singleForm reducer. 
Static dummy data from JSON removed.

**_FRONTEND CHANGES(10)_**
`services/ui-src/src/scss/_form.scss`
Styling for the form 

`services/ui-src/src/utilityFunctions/sortingFunctions.js`
I made a file for utility functions we'll be using on the frontend. 
This function is very similar to the `sortQuestionsByNumber` in what it does but this one is parsed in the component instead of in redux because the array its sorting is MASSIVE. By the time it reaches the front end  the array has already been narrowed down a bit. 
sortQuestionColumns takes in the array of `rows` from answers data. The GridWithTotals component needs the rows to be in order for it to render properly. sortQuestionColumns grabs the last digit of each column name and orders the object in descending order by column name (col1... col6). sortQuestionColumns takes in just a single question's `rows` array at a time, sorts them and returns the array. 

`services/ui-src/src/components/CertificationTab.js`
`services/ui-src/src/containers/NotFound.js`
The shape of the currentForm reducer changed to include questions, answers and status data as separate objects.
Any component previously accessing that reducer in this way state.currentForm would now have to access it through one more layer ie:  (state.currentForm.statusData). In the files listed above, this was the only change. 

`services/ui-src/src/Routes.js`
Added FormPage component to the router as an AuthenticatedRoute. 

`services/ui-src/src/components/GridWithTotals/GridWithTotals.js`
Only two changes were made to the GridWithTotals component. 
1- The demo caption ( caption="This is an example of a table with totals") was removed. 
2- each of the text fields that take in input were given a "disabled" property that will be sent to the component via props. The default property is "undefined" so its absence will break nothing. When a form is meant to be disabled (form is certified & submitted, or form does not apply), by passing down this prop (as a boolean) users will be unable to edit the form. 

Its also worth noting here that Question.js, the parent component of GridWithTotals, is passing down a "questionID" prop. Each GridWithTotals is now associated with a specific `answer_entry` ID, so state/year/quarter/form/age range/ question number. GridWithTotals is not currently doing anything with this data, but it is available to it via props. 

`services/ui-src/src/components/form/FormPage.js`
Form page is fetching all of the data needed for the components on the form page and updating redux with the data fetched so each component that is connected can fetch data from there. 
the state/year/quarter and form are extracted from the URL. Those parameters are formatted and useEffect triggers the Thunk getForm(). getForm() updates the "currentform" reducer. 

The JSX renders the FormHeader, the TabContainer and the FormFooter. The header and footer require similar data so its being passed directly to them via props so there is no need for them to be redux connected. 

`services/ui-src/src/components/layout/TabContainer.js`
The TabContainer is where the questions will sit. They're split between tabs(age ranges). The tabs are being mapped through. 
For each tab (age range): a range_id is extracted and from there the answers array is filtered to return the answers just for that tab (age range). 
For each question: The ID is extracted from each question, the corresponding answer object is found in the answers array. The question component is passed the ID for the age range its in, questionData for the label and questionID, answerData for the answerID and the rows they're meant to be stored in and a disabled prop. 

The disabled prop is based on the current form's "not_applicable" and "status" properties. 

`services/ui-src/src/components/form/FormHeader.js`
This component is for displaying everything above the tab component. This component is passed all of its props from FormPage.js

`services/ui-src/src/components/form/FormFooter.js`
This component is for displaying everything below the tab component. This component is passed all of its props from FormPage.js

`services/ui-src/src/components/Question.js`
Question receives one question at a time. The label and question ID are taken from the question object in Formpage.js
The rows that house the answers and the answer ID (answer_entry) from the answer object in Formpage.js. 
The array of rows is sorted with a helper function. The rows must be ordered before being passed to GridWithTotals. 
GridWithTotals is given the props questionID (an ID specific to a question in a given year/state/quarter/form/age range), gridData (the rows from the answer JSON) and disabled (a boolean determining if the form should be disabled). 

**_BACKEND CHANGES(3)_**
`services/app-api/serverless.yml`
Added to the list of API functions for the new endpoint. 

`services/app-api/handlers/forms/get/getFormTypes.js`
Handler for the new endpoint. It was super straightforward as I'm returning the entire table. The local form table only has 6 small entries in total. If we want to get more specific, i'd be happy to adjust the query so that we're fetching only the one we need. 

`services/ui-src/src/libs/api.js`
Function for calling the new endpoint 

**AC:**
- [x] Selected form displays completely (Title/Name, Narrative, Age Range Tabs, Questions, Answers)
- [x] User can navigate through tabs and enter data into each
- [x] Links take user to appropriate places

**Notes:**
The ageRanges for the tabs is still coming from static data because there is no endpoint for it. We should make an endpoint for it and update the global.age_ranges. This could become more complicated as the states can alter their age ranges, but "as-is" it would be a very small ticket. 1-2 points. 

Forms being disabled needs to be looked at. Maybe the "status" property in the status table should be shortened? In any case, when the ticket for disabling the form is being worked on the 'disabled' boolean in mapState() in TabContainer.js should be adjusted! That is where the boolean is being determined and then passed to all of the children components from.


alewoo28:
The 9 codeclimate "issues" are nonsense.
The first three are about proptypes being too similar. PropTypes objects will always be similar by design. 

Two complains are about the FormHeader and FormFooter components being similar. Theres nothing to fix there, they are similar. Making them into one component would be a mistake. 

Four complaints are about GridWithTotals, refactoring that component is outside the scope of this ticket and I dont think the "issues" are even pertinent.

cl-adavyd (approved):

StephanieCoppel:
approved

eolaniyan052 (approved):

jgillis1:
Testing Steps:

Set your user state a “state” user with AL, MD or PA as its state

In local/AWS, update the login information for Alice in LocalLogin.js

In DEV, change your information in master’s user_auth table

From the Home Page, click on 2021 > Quarter 1

From the Quarter Page, click on the link for 21E

In the form you should be able to navigate each age range tab and view the Summary (separate ticket) and Certification tabs

Data is not currently saving but it can be entered.

PR #172, opened 2021-03-17
Fix 21PW Questions

jgillis1:
**Purpose**
There was no age range populating in the question label, it was saying "undefined". Fix the question label so that it reads properly and does not show "undefined" when trying to replace the &&&VARIABLE&&& (which is not actually needed for most of these questions).

**Files Changed**
- 21PW.json
- 21PW_2020.json


TGriesemer (approved):
Looks good!

mblevins-collabralink (approved):
I'm curious why the variables were in there in the first place?

PR #173, opened 2021-03-17
Feature synthesized component

TGriesemer:
For ticket [6531](https://qmacbis.atlassian.net/secure/RapidBoard.jspa?rapidView=184&modal=detail&selectedIssue=OY2-6531)

This component (question 5 on forms 21E, 21PW, 64.21E, and 64.EC) calculates a value based on its formula (<0> / <1> so far) and targets. The targets point to another question/value(Q4 and Q1 so far) in the same age range and the formula gives direction on how to complete the calculation.

TESTING: (NOTE: All values should be rounded to 1 decimal place)
If you want to use existing values:
1 Navigate to /forms/PA/2021/1/64-21E (only the first two age ranges have data)
2 Question number 5 will be the calculated value of (Q4 / Q1) from their respective position (Q5.rows[1].col2 = Q4.rows[1].col2/Q1.rows[1].col2)

If you want to create your own test:
1. Update Q1 and Q4 for any of the following forms (21E, 21PW, 64.21E, and 64.EC). 
2. In the same age range, Q5 should have the calculated values of Q4/Q1

### DIFF
--- a/services/ui-src/src/components/GridWithTotals/GridWithTotals.js
+++ b/services/ui-src/src/components/GridWithTotals/GridWithTotals.js
@@ -16,6 +16,14 @@ const GridWithTotals = props => {
     updateTotals();
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+  const roundValue = (value) => {
+    let returnValue = value;
+    if(value % 1 > 0){
+      returnValue = (Math.round(value * 100) / 100).toFixed(1);
+    }
+    return returnValue;
+  }
+
   const updateGrid = (row, column, event) => {
     let gridCopy = [...gridData];
     gridCopy[row][column] = parseFloat(event.target.value);
@@ -138,7 +146,7 @@ const GridWithTotals = props => {
                       onChange={event =>
                         updateGrid(rowIndex, columnIndex, event)
                       }
-                      defaultValue={column}
+                      defaultValue={roundValue(column)}
                       disabled={props.disabled}
                     />
                   </td>
@@ -151,7 +159,7 @@ const GridWithTotals = props => {
                     type="number"
                     className="grid-column"
                     onChange={event => updateGrid(rowIndex, columnIndex, event)}
-                    defaultValue={column}
+                    defaultValue={roundValue(column)}
                     disabled={props.disabled}
                   />
                 </td>
@@ -160,7 +168,7 @@ const GridWithTotals = props => {
 
             return formattedCell;
           })}
-          <td className="total-column">{gridRowTotals[rowIndex]}</td>
+          <td className="total-column">{roundValue(gridRowTotals[rowIndex])}</td>
         </tr>
       );
     }
@@ -178,11 +186,13 @@ const GridWithTotals = props => {
         </th>
       );
     } else {
-      column = <td className="total-column">{gridColumnTotals[i]}</td>;
+      column = <td className="total-column">{roundValue(gridColumnTotals[i])}</td>;
     }
 
     return column;
   });
+
+
 
   return (
     <div className="grid-with-totals">
@@ -194,7 +204,7 @@ const GridWithTotals = props => {
           {tableData}
           <tr>
             {totalsRow}
-            <td className="total-column">{gridTotalOfTotals}</td>
+            <td className="total-column">{roundValue(gridTotalOfTotals)}</td>
           </tr>
         </tbody>
       </Table>
--- a/services/ui-src/src/components/SynthesizedGrid/SynthesizedGrid.js
+++ b/services/ui-src/src/components/SynthesizedGrid/SynthesizedGrid.js
@@ -2,20 +2,18 @@ import React, { useEffect, useState } from "react";
 import GridWithTotals from "../GridWithTotals/GridWithTotals"
 import PropTypes from "prop-types";
 import {connect} from "react-redux";
-import jsonpath,{selectObjectInArrayByQuestionId} from "../../utilityFunctions/jsonPath";
+import {selectObjectInArrayByQuestionId} from "../../utilityFunctions/jsonPath";
 import { sortQuestionColumns } from "../../utilityFunctions/sortingFunctions";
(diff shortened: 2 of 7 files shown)


mblevins-collabralink (changes requested):
I think this needs another set of eyes more familiar with the exact goals, but I also think there's some refactoring to be done here.

services/ui-src/src/components/GridWithTotals/GridWithTotals.js
```diff
@@ -16,6 +16,14 @@ const GridWithTotals = props => {
     updateTotals();
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+  const roundValue = value => {
```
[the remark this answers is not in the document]
TGriesemer: It checks to see if the value contains a decimal or not. If it does not contain one, then don't display a decimal. If we don't use this and round all values that are returned, then every box contains a decimal. If we don't try to round all values that are returned, we need to check every value to see if it needs to be rounded or not.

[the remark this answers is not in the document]
mblevins-collabralink: The cycle-minded guy in me wants to do this:
```

    if (value % 1 !== 0) {
      return value.toFixed(1);
    }
    return value;
```
This works in the console, and should be more efficient. Neither one behaves as desired with something like 1.99 (which, of course, rounds to 2.0) as a value, but I don't know if you're getting 2 digit decimals or not.

This may seem nitpicky, but this is easier to read and understand than what you had before, and doesn't introduce variables you don't need. It's very obvious what it's doing.

mblevins-collabralink (changes requested):

services/ui-src/src/components/GridWithTotals/GridWithTotals.js
cl-adavyd, line 19: I don't think this roundValue function should be contained in this component.  If it's a formatting function, and adds some functionality to this component that we can use somewhere else, it should be contained elsewhere and imported into the component for the purposes of use, but nothing about it ties it to this component specifically.

cl-adavyd: additionally, I'm not sure formatting should be conditional based on whether or not the value contains decimal points.  it is either rounded or isn't.  Precision for all numerical data is usually the same at lest throughout the component

TGriesemer: I can definitely move it to a new file. The reasons I put it in this component was because, as far as I know, it is the only component that will be handling values with a decimal for this entire project. 

The reason the rounding is a conditional is because either we are talking about a whole number, which doesn't need to be rounded and also doesn't need to display a decimal, or we are talking about a calculated value, which will most likely result in a value with several decimal places and needs to be rounded.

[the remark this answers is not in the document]
TGriesemer: This looks like a good simple refactor!

cl-adavyd: but rounding a whole number will not result in an error.  and it will get rid of the need to have a conditional.  

as far as decimal precision, I believe 1 digit is called for.  it is normally applied to all numbers being displayed unless explicitly called for otherwise.

I understand that for right now, as far as we know this rounding takes place in only this component, but logically this functionality is not tied to this component ONLY, and should not be included in it.  it is a formatting function on something this component provides

TGriesemer: I will remove this function from this component and move it to a proper location. 

The reason that the conditional is there is so that when we are handed a whole number, it doesn't try to round it and add an unnecessary decimal point. I think it looks better without every whole number being listed with a __.0. If you think that is is not needed then I can remove it.

services/ui-src/src/components/SynthesizedGrid/SynthesizedGrid.js
```diff
@@ -0,0 +1,104 @@
...
+      Object.entries(incomingCalculation.targets).forEach(key => {
+        tempCalculation[key[0]] = getValue(key[1], tabAnswers);
+      });
+      // calculates the value based off of the formula <0> / <1>
+      returnValue = tempCalculation[0] / tempCalculation[1];
+      return returnValue;
+    }
+  };
+
+  const getValue = (target, tabAnswers) => {
```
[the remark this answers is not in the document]
TGriesemer: Does anyone have any suggestions on how this might be improved or be written so it is reusable?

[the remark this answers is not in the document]
mblevins-collabralink: Abstracting it would make it better. You're looking for a specific row ([2]) in the array. If the data format changes, you have to rewrite the code, rather than the call. If you pass the headerrow as a parameter, for example, it would be less fragile. Tying it a specific data format like that makes it hard to reuse.

services/ui-src/src/components/GridWithTotals/GridWithTotals.js
cl-adavyd: as previously stated, numbers are normally handled in a similar fashion regardless of their format, unless very explicitly called for otherwise.  it is not a matter of personal preference, but more an engineering standard.

if numerical output is rounded and displayed, it should be handled similarly for the entirety of said numerical output.  at least as far as the data set in question is concerned

services/ui-src/src/utilityFunctions/mathFunctions.js
```diff
@@ -1,6 +1,6 @@
 export const roundValue = value => {
```
cl-adavyd, line 1: this function is not needed.  the same thing can be accomplished inline without a conditional.

example:

Math.round(value).toFixed(currentPrecision)

where currentPrecision = 1, for our case

TGriesemer: The rounding function has been replaced with the suggested code.

PR #6798, opened 2023-04-12
Force scan to consider last evaluated, and remove dangerous alternative

BearHanded:
### Description
Update the scan to consider last evaluated.
Removed a dangerous recursive one that tracked with a global list of items.

I'm on the fence as to whether or not the recursive search should be opt-in or just for all, but I think this bug proves that not having default for all can get devs into some risky spots. Future improvement might be to make the scan optionally opt-out of recursive search, but I have no use case for it at the moment.

You should see this when you try to create existing data now:
image

### Related ticket(s)

MDCT-2400

---
### How to test

Pull it down, and interact with the app. Nothing should break. Generate a lot of forms (2.5MB). You shouldn't be able to overwrite existing data.

### Important updates

---
### Author checklist

- [ ] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
---

### DIFF
--- a/services/app-api/handlers/forms/post/obtainAvailableForms.js
+++ b/services/app-api/handlers/forms/post/obtainAvailableForms.js
@@ -1,5 +1,5 @@
+import dynamodbLib from "../../../libs/dynamodb-lib";
 import handler from "../../../libs/handler-lib";
-import { recursiveScan } from "../../shared/sharedFunctions";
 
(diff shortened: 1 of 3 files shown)


braxex (approved):

ntsummers1:
Confirmed this works locally as well

PR #12916, opened 2023-04-24
Bugfix totals calc

ntsummers1:
### Description

The bottom row still had some slight errors when it came to totals. I went and made those changes and refactored some things slighty. Could be much better (The refactoring part) but we'll take this for now.

### Related ticket(s)

MDCT-481, MDCT-2390

---
### How to test

Load up 64.21E. Fill some numbers out, see it gives the expected answer!

### Important updates

---
### Author checklist

- [x] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
---

### DIFF
--- a/services/ui-src/src/components/SynthesizedGrid/SynthesizedGrid.js
+++ b/services/ui-src/src/components/SynthesizedGrid/SynthesizedGrid.js
@@ -23,7 +23,7 @@ export const SynthesizedGrid = ({ entireForm, range }) => {
     return true;
   });
 
-  let numOr0 = n => (isNaN(n) ? 0 : n);
+  let numOr0 = n => (isNaN(n) ? 0 : Number(n));
 
   let firstRowQ4Total = Object.values(q4arry[0].rows[1]).reduce(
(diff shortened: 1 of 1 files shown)


braxex (approved):
LBTGTM

benmartin-coforma:
I am not jealous of this task. My suggestion is to convert the data into what you want it to be, before doing any math. Get you a nice 2d array of `number`s, maybe like this:

```js
  const numericQuestionAnswers = (questionFilterString) => {
    return tabAnswers.find((e) => {
        return e.question.includes(questionFilterString)
      })!.rows
      .map(row => Object.values(row)
        .map(Number)
        .map(n => isNaN(n) ? 0 : n));
  }

  let question4Answers = numericQuestionAnswers("-04");
  let question1Answers = numericQuestionAnswers("-01");
```

Perhaps also a "divide if finite, then format" function, to save time in the markup:
```js
  const divideAndFormat = (a, b) => (b === 0 ? 0 : a / b).toFixed(1);
```

But then, both of these would mean a pretty thorough overhaul of the page, which may be something you're trying to avoid. So it's up to you.

PR #14673, opened 2023-07-24
remove okta metadata from github secrets; conditional idp

gmrabian:
### Description
Dependabot deploys were failing due to dependabot not having permissions to access the OKTA_METADATA_URL GitHub secret. Whereas we could have permitted it to view that secret, we determined rather to align the serverless behavior in SEDS a little closer with the other MDCT applications. This is also gets us closer to having user roles auto-generated rather than self signup.

Once this merges we no longer need the OKTA_METADATA_URL GitHub secret

If we are in a none-trunk branch (master, val, production) then it will not even provision the identity provider backing

The necessary ssm params already exist in dev, val, and prod
Screenshot 2023-07-24 at 2 46 11 PM
Screenshot 2023-07-24 at 2 47 51 PM
Screenshot 2023-07-24 at 2 48 36 PM

### Related ticket(s)
MDCT-2533

---
### How to test
View [dependabot PR](https://github.com/Enterprise-CMCS/macpro-mdct-seds/pull/14672) that I tested this on:
In the first action (upon creation) it [failed to deploy](https://github.com/Enterprise-CMCS/macpro-mdct-seds/actions/runs/5599259726/job/15167035592) with this error:
`Cannot resolve variable at "custom.okta_metadata_url": Value not found at "env" source`
Because it could not access that env variable

After adding the changes seen in this PR it [succeeded in deploying](https://github.com/Enterprise-CMCS/macpro-mdct-seds/actions/runs/5646977819/job/15296083340) 

### Next steps
- Merge `master` into dependabot branches so they rerun their actions with this setup
- Make sure they pass
- Approve them and they will auto-merge!

---
### Author checklist

- [x] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
---


britt-mo (approved):

BearHanded (approved):
👍

PR #14729, opened 2023-09-08
Remove package duplicates in ui-src

gmrabian:
### Description
While reading numerous deploy logs I noticed we had two packages that existed in `devDependencies` and `dependencies`
There was also no license in the ui-src package

Screenshot 2023-09-08 at 3 18 21 PM

---
### How to test

ui-src builds correctly


services/ui-src/package.json
```diff
@@ -192,9 +192,8 @@
     "@babel/plugin-transform-react-jsx": "^7.12.17",
     "core-js": "^3.9.1",
     "eslint-config-prettier": "^7.1.0",
-    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-jsx-a11y": "^6.7.1",
```
gmrabian, line 195: left final eslint in devDependencies because eslint should not be in the deployed build

services/ui-src/package.json
```diff
@@ -39,15 +40,14 @@
...
-    "eslint-plugin-jsx-a11y": "6.2.3",
     "eslint-plugin-react": "7.19.0",
     "eslint-plugin-react-hooks": "^1.6.1",
     "file-loader": "4.3.0",
     "file-saver": "^2.0.5",
     "fs-extra": "^8.1.0",
     "html-webpack-plugin": "^4.5.2",
     "html2canvas": "^1.0.0-rc.7",
-    "identity-obj-proxy": "3.0.0",
+    "identity-obj-proxy": "^3.0.0",
```
gmrabian, line 50: left this in the build dependencies because I wasn't exactly sure where it should end up. I think it's only used in jest

gmrabian (commented):
some notes

services/ui-src/package.json
gmrabian: same for identity-obj-proxy

BearHanded (approved):

PR #14759, opened 2023-10-30
Add security headers

benmartin-coforma:
### Description
This PR adds several HTTP headers when serving the website:
* One which instructs browsers _only_ to interact with us via HTTPS
* One which instructs browsers not to serve our content within an iframe

The structure of the function to add these headers is identical to that of our other 4 apps: CARTS, QMR, MCR, MFP

### Related ticket(s)
CMDCT-271, CMDCT-308

---
### How to test
We don't use iframes to serve this website, so this PR shouldn't change anything. A quick smoke test (as provided by Cypress) should be sufficient.

It will also be good to inspect the network tab when loading the site for the first time. The new headers should be present on the response.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui/serverless.yml
+++ b/services/ui/serverless.yml
@@ -206,6 +206,9 @@ resources:
               Cookies:
                 Forward: none
             ViewerProtocolPolicy: redirect-to-https
+            FunctionAssociations:
+              - EventType: viewer-response
+                FunctionARN: !GetAtt HstsCloudfrontFunction.FunctionMetadata.FunctionARN
           ViewerCertificate:
             Fn::If:
               - CreateCustomCloudFrontDomain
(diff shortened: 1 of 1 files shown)


karla-vm (dismissed):

services/ui/serverless.yml
```diff
@@ -232,6 +235,22 @@ resources:
...
+            var response = event.response;
+            var headers = response.headers;
+            headers['strict-transport-security'] = { value: 'max-age=63072000; includeSubdomains; preload'};
+            headers['x-frame-options'] = { value: 'DENY' };
+            return response;
+          }
+        FunctionConfig:
+          Comment: This function adds headers to implement HSTS
+          Runtime: cloudfront-js-1.0
+        Name: hsts-${self:custom.stage}
```
gmrabian, line 253: want to change these to not be HSTS?

benmartin-coforma: Yes I do!

gmrabian (approved):

PR #14760, opened 2023-10-30
Upgrade uswds to 2.14

benmartin-coforma:
### Description
This PR bumps the version of the `uswds` package. The previous version had a vulnerability. A few notes:
* This app uses `@trussworks/react-uswds` rather than `uswds` directly, mostly. It is unlikely we had any exposure to the vulnerability.
* This app _does_ include some CSS from `uswds` directly. This seems redundant; surely the `@trussworks` package includes all relevant CSS already? But I believe that proving this would be difficult.
* The `uswds` package is not namespaced, but there does exist a namespaced version: `@uswds/uswds`. Possibly the latter is meant to supersede the former? Perhaps if we are upgrading packages we should be switching to the namespaced version? But I'm not 100% sure that's the case. And in any case, I would prefer to remove this dependency altogether, so that we would wholly rely on `@trussworks`.

So I'm only bumping the version number here.

### Related ticket(s)
CMDCT-2854

---
### How to test
I guess make sure the site looks OK? In theory, it's possible that this PR changes some of the CSS.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/package.json
+++ b/services/ui-src/package.json
@@ -98,7 +98,7 @@
     "ts-pnp": "1.1.6",
     "typescript": "^4.2.4",
     "url-loader": "2.3.0",
-    "uswds": "^2.11.1",
+    "uswds": "^2.14.0",
     "webpack": "4.42.0",
     "webpack-dev-server": "^3.11.3",
     "webpack-manifest-plugin": "2.2.0",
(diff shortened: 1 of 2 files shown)


gmrabian (approved):
seems ok; deployed branches are not entirely testable anyway

BearHanded (approved):

PR #14794, opened 2024-01-31
CMDCT-3185 + CMDCT-3186 - Auth + Last Edited

benmartin-coforma:
### Description

PR brought to you by the Ben-Brendan-Karla hivemind.

- Adds the cognito JWT to all requests to the API from the UI
- Uses that to first initate user permissions before reading any changes from the users table
- API endpoints check for permissions and relevant state.
- Notable only admins can call updateUser, closing the finding for privilege escalation EXCEPT when a user is modifying their empty state list the first time.

### Related ticket(s)

CMDCT-3185
CMDCT-3186

---
### How to test

- The app should continue to function as normal
- Check for privilege escalation

### Important updates

---
### Author checklist

---


services/app-api/handlers/forms/post/generateQuarterForms.js
```diff
@@ -22,6 +23,8 @@ export const main = handler(async (event, context) => {
     return null;
   }
 
+  await authorizeAdmin(event);
```
BearHanded, line 26: This will break the quarterly job without the other changes _but_ that also should be a bit out there until it needs to trigger.

We either need the alternate handler or agree we're following up with the other change quickly.

benmartin-coforma: I think an alternate handler is the ideal solution. It feels odd to me to have a single lambda that handles two different types of event.

services/ui-src/src/components/App/App.js
```diff
@@ -23,14 +20,10 @@ function App() {
...
-      // *** make sure attributes exist and are in standard format
-      user.attributes = user.payload;
-      user.attributes["app-role"] = await determineRole(
-        user.attributes["custom:ismemberof"]
-      );
-
-      await ascertainUserPresence(user);
+      const token = (await Auth.currentSession()).getIdToken();
+      const apiUser = await ensureUserExistsInApi(token.payload.email);
+      const user = { attributes: apiUser }; // ew
```
gmrabian, line 25: ew?

gmrabian (approved):
pretty slick!

ajaitasaini (approved):
🙌 amazing

services/ui-src/src/components/App/App.js
benmartin-coforma: Just to explain myself here... I would expect a user object to resemble this:
```typescript
user = {
  username: "george",
  email: "george@example.com",
  roles: "state_user"
}
```

Instead, we have this:
```typescript
user = {
  attributes: {
    username: "george",
    email: "george@example.com",
    roles: "state_user"
  }
}
```

This is for no reason that I can tell, and it adds noise throughout ui-src whenever we need to access those attributes. I find it gross, but to fix it would mean a whole _nother_ raft of changes. Maybe someday.

PR #14796, opened 2024-02-01
Fix snykdependa destroy

berryd:
### Description
This PR fixes two specific issues:

1. The Snyk integration implemented a change which was applied to the deploy action, which shortens auto-generated branches to a consumable length.  Dependabot already had a case in place to do this, but the change altered how Dependabot branches were interpolated into stage names, thus breaking the ability of the destroy action to correctly match infrastructure with a branch.  This caused infrastructure to leak.
2. Destroy actions did not fail when no infrastructure was identified.  Therefore successfully executed destroy actions did not indicate whether or not the action successfully destroyed the associated infrastructure.

There is a bit of a conundrum with this PR, in that it can't be tested in full until it's merged into master.   This is because destruction of a branch triggers destroy to run on the default branch, which is generally `master` or `main`.  Each component was tested individually, but the full CI run cannot be validated until this is merged in.  

### Related ticket(s)

CMDCT-3218

---
### How to test
Each component has been validated individually, but should be tested in entirety once this is merged in.

Branch from `master`, and create three branches to exercise every condition.
`dependabot/npm_and_yarn/berry-did-this` (will be interpolated to stage as `x9e7b964750 `)
`snyk-upgrade-berry-did-this9faf06c882b46cc4a8e83` (will be interpolated to stage as `sthis9faf06`)
`berry-actual-change` (will be the same stage as branch name)

Let the builds go through, then delete the branches
Verify the infrastructure is successfully destroyed

### Important updates

This change will need to be propagated to every other product once verified.

---
### Author checklist

---

### DIFF
--- a/.github/setBranchName.sh
+++ b/.github/setBranchName.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+GITHUB_REF="${1}"
+
+[ -z "${GITHUB_REF}" ] && echo "Error setting branch name.  No input given." && exit 1
+
+case ${GITHUB_REF} in
+  $([[ "$GITHUB_REF" =~ ^refs/heads/dependabot/.* ]] && echo ${GITHUB_REF}))
+    echo ${GITHUB_REF##*/*-} | md5sum | head -c 10 | sed 's/^/x/'
+    ;;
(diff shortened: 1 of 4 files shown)


BearHanded (approved):

benmartin-coforma (approved):
The shell scripting is not within my expertise, but it's a good idea and from what I can tell it looks sound.

PR #14802, opened 2024-02-05
Init

berryd:
### Description
A previous pull request functionally satiated the requirements for the associated ticket, but while debugging an issue with destroy, some discrepancies with references and look ups were discovered.

- `GITHUB_REF` environment variable is the same as `github.ref` and includes the fully qualified path
- Destroy actions always trigger on the default branch so `GITHUB_REF`/`github.ref` would be `/refs/heads/master`
- Destroy actions know what the target is based on `github.event.ref` which is NOT the fully qualified path
- Defining `workflow_dispatch` enables manual triggering destroy actions against a given code base and enables specifying target.

The changes in this pull request do the following:
- Apply consistent lookups in Deploy (concurrency lock would be deploy-[branch] rather than deploy-/refs/heads/[branch])
- Remove hard coded addition to destroy action which would break if a tag is deleted
- Enable manual dispatching of events.

For other repositories which may differ in structure, it is suggested that the `workflow_dispatch` be added as an initial pull request, as this has to make it to the default stage before it can be acted upon.  This will enable development against destroy actions which can only trigger from the default branch until the configuration change is merged.

### Related ticket(s)

CMDCT-3218

---
### How to test
Once this has been merged...

Create three branches
- dependabot/npm_and_yarn/berry-did-this
- snyk-upgrade-berry-did-this9faf06c882b46cc4a8e83
- berry-test

Push and wait for builds to succeed.

Delete the remote branches and verify the destroy step completes.  

Create one more branch, any arbitrary name, and add `echo this is a test && exit 1` to the destroy yaml file at line 37.  Exercise the manual trigger and verify the update in the unmerged branch is executed in the destroy step

### Important updates

---
### Author checklist

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ on:
       - "!skipci*"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref_name }}
 
 permissions:
   id-token: write
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: set branch_name # Some integrations (Dependabot & Snyk) build very long branch names. This is a switch to make long branch names shorter.
         run: |
-          BRANCH_NAME=$(./.github/setBranchName.sh ${GITHUB_REF})
+          BRANCH_NAME=$(./.github/setBranchName.sh ${{ github.ref_name }})
           echo "branch_name=${BRANCH_NAME}" >> $GITHUB_ENV
       - name: Validate branch name
         run: ./.github/branchNameValidation.sh $STAGE_PREFIX$branch_name
@@ -78,7 +78,7 @@ jobs:
   e2e-tests:
     name: E2E Integration Tests (Cypress)
     needs: deploy
-    if: ${{ github.ref != 'refs/heads/production' }}
+    if: ${{ github.ref_name != 'production' }}
     runs-on: ubuntu-latest
     container:
       image: cypress/browsers:node16.16.0-chrome107-ff107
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -8,7 +8,7 @@ on:
       - "!skipci*"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref_name }}
```
berryd, line 11: github.ref is the same as GITHUB_REF and fully qualifies the path, so a branch `berrydidthis` would appear as `/refs/heads/berrydidthis` hence the need to parse out the strings.

In destroy, GITHUB_REF and github.ref will always point to `master` which is the default branch on which destroy is executed.  In that case, we need to get the reference from the event.

What annoys me is that github.ref_name and github.event.ref are both just the branch without the full qualification.  I would prefer a consistent naming convention but this is out of my hands.

.github/workflows/destroy.yml
```diff
@@ -1,6 +1,12 @@
 name: Destroy
 
-on: delete
+on:
+  delete:
+  workflow_dispatch:
```
berryd, line 5: This will enable the manual execution of destroy against any given stage without needing to locate the stage to destroy it.  It maintains the checks that prevent destruction of any of the integration states.  One small issue though, is that again, this has to execute from the default branch, so this has to be merged in to test.

BearHanded (approved):

.github/workflows/deploy.yml
```diff
@@ -78,7 +78,7 @@ jobs:
   e2e-tests:
     name: E2E Integration Tests (Cypress)
     needs: deploy
-    if: ${{ github.ref != 'refs/heads/production' }}
+    if: ${{ github.ref_name != 'production' }}
```
gmrabian, line 81: i like this

gmrabian (approved):

PR #14833, opened 2024-03-22
[CMDCT-3440] Tealium Fixes

BearHanded:
### Description

Update the logic around when to make tealium page view calls.
- The page would sometimes not report at all at /login. Race condition with retrieving and attaching the tealium script. Add simple retry logic X times to see if the script is attached, otherwise bail.
- The page double reported when logged in. The hook needs to only fire once loading the user permisions into memory settles. 
- Some extra logic to not double report the /login page on redirect from /
- Corrects the url param used when in production from `production` to `prod`. This is not the same as the site_environment flag, and the two have been flipped back and forth together in the past.

### Related ticket(s)

CMDCT-3440

---
### How to test

- Add breakpoints to the tealium library, or the calls in tealium.js before the call goes out in `window.utag.view`, OR add logging when running locally to tealium.js.
- Each page view should only trigger one report.
- Pay extra attention to the login page and root url. The root url redirects to login and needed some massaging to only send once.

Happy to show how to test this better if necessary.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/public/index.html
+++ b/services/ui-src/public/index.html
@@ -59,7 +59,7 @@
     <!-- Tealium Analytics Tag Manager -->
     <script>
       var tealiumEnvMap = {
-        "mdctseds.cms.gov": "production",
+        "mdctseds.cms.gov": "prod",
         "mdctsedsval.cms.gov": "qa"
       };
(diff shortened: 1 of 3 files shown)


benmartin-coforma (approved):
The logic seems sound to me.

ailZhou (approved):

PR #14846, opened 2024-04-03
Upgrade to AWS SDK V3

gmrabian:
### Description

Upgrade to [AWS SDK V3](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/).

- Upgrade all libs to V3
- Update handlers as necessary 
  - notification handlers using SES
  - remove looping from handlers where it's handled by the lib

Note: the `app-api/` folder had zero tests before this upgrade. this introduces a high risk when making significant changes. i chose to not add tests as part of this PR due to 1. size, 2. scope

if you think i should add tests, particularly for things like the libs, let me know!

### Related ticket(s)

CMDCT-3073

---
### How to test

- SEDS still works!

### Notes

- Will need to run `yarn` in your `app-api/` directory

---
### Pre-review checklist

- [ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary -- SEE NOTES ABOVE


services/app-api/libs/dynamodb-lib.js
```diff
@@ -1,55 +1,102 @@
...
 
-const client = new AWS.DynamoDB.DocumentClient(dyanmoConfig);
+    errorMessage += `\n  ${tableName}: ${failedCount} (of ${attemptedCount}) requests failed`;
+  }
+
+  return errorMessage;
+};
 
 export default {
-  get: (params) => client.get(params).promise(),
```
gmrabian, line 20: get wasn't used anywhere

services/app-api/libs/dynamodb-lib.js
```diff
@@ -1,55 +1,102 @@
...
 
-      params.ExclusiveStartKey = result.LastEvaluatedKey;
-      complete = result.LastEvaluatedKey === undefined;
-    }
+      ExclusiveStartKey = result.LastEvaluatedKey;
+    } while (ExclusiveStartKey)
     return itemSet;
   },
-  batchWrite: (params) => client.batchWrite(params).promise(),
-  listTables: (params) => database.listTables(params).promise(),
```
gmrabian, line 54: listTables wasn't used anywhere

services/app-api/libs/kafka-source-lib.js
```diff
@@ -40,7 +40,7 @@ class KafkaSourceLib {
   */
 
   unmarshallOptions = {
-    convertEmptyValues: true,
```
gmrabian, line 43: this options does not exist in the new version; not sure if it fails us along the way

services/app-api/libs/kafka-source-lib.js
```diff
@@ -40,7 +40,7 @@ class KafkaSourceLib {
...
+    convertWithoutMapWrapper: false,
```
gmrabian, line 43: this should be false by default but I'm just making it explicit what our options are

services/app-api/libs/kafka-source-lib.js
benmartin-coforma: I don't think it ever existed; this is a marshalling option, not an unmarshalling one. I don't believe it caused any negative effects... it would have just been ignored.

services/app-api/handlers/kafka/get/forceKafkaSync.js
```diff
@@ -78,34 +64,25 @@ export const main = handler(async (event, context) => {
...
-      } catch (err) {
-        console.error(`Database scan failed for the table ${tableName}
-                     with startingKey ${startingKey} and the keepSearching flag is ${keepSearching}.
-                     Error: ${err}`);
-        throw err;
-      }
+    try {
+      data = await dynamoDb.scan({
+        TableName: tableName,
+      });
```
benmartin-coforma, line 72: This is actually the one case (across all of our apps, as far as I know) where the scan pagination was relevant. The `mergeLastSynced` operation can run on each page of results while we wait for the next, and processed objects can be garbage collected as the data streams through this lambda. And, if I recall correctly, SEDS has a _ton_ of data in dynamo, to the point where that might start to matter.

If we want to preserve those performance characteristics... the easiest thing would be, basically, to leave this file alone. Perhaps the legacy `scan` function would remain in dynamodb-lib for now, under the scarier name `scanSome`?

I'd love to be able to use the SDKv3 `paginate` here, but it's locked behind a somewhat intimidating .tsconfig change.

services/app-api/libs/dynamodb-lib.js
```diff
@@ -1,55 +1,71 @@
...
+    do {
+      const command = new ScanCommand({ ...params, ExclusiveStartKey });
+      const result = await client.send(command);
+      items = items.concat(result.Items ?? []);
+      ExclusiveStartKey = result.LastEvaluatedKey;
+    } while (ExclusiveStartKey);
+
     return { Items: items, Count: items.length };
   },
   scanMapToSet: async (params, mapFunction) => {
```
benmartin-coforma, line 52: I would like to see this function deleted. If it doesn't feel like too much scope creep, would you consider rewriting the one call site, in sharedFunctions.js? I think it could be as simple as:
```ts
  const result = await dynamoDb.scan(params);
  return new Set(result.Items.map((answer) => answer.state_form));
```

services/app-api/handlers/kafka/get/forceKafkaSync.js
gmrabian: The old `scan` would have performed a full scan as well, despite the pagination logic expected in `forceKafkaSync`. We will keep the same full scan logic for now, knowing it can be improved later.
https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/master/services/app-api/libs/dynamodb-lib.js#L25-L35

benmartin-coforma (approved):
Very nice work.

BearHanded (approved):

PR #14859, opened 2024-04-22
Improve Artifact Upload

berryd:
### Description
The change here are smaller than in the newer products, as there are not multiple test suites to produce artifact file collisions, but the same change have been introduced to improve consistency.  

### Related ticket(s)

CMDCT-3555

---
### How to test

I pushed a build set to always upload artifacts.  The artifacts can be seen to be correctly publishing [here](https://github.com/Enterprise-CMCS/macpro-mdct-seds/actions/runs/8789118010).  A successful build will not produce artifacts.

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -197,6 +197,8 @@ jobs:
           working-directory: tests/cypress
           spec: integration/**/*
           browser: chrome
+      - name: Debug directory structure
+        run: apt-get install -y tree && tree
       - name: Upload screenshots
         uses: actions/upload-artifact@v4
         if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -173,9 +180,6 @@ jobs:
       - register-runner
     if: ${{ always() && !cancelled() && needs.deploy.result == 'success' && github.ref_name != 'production' }}
     runs-on: ubuntu-latest
-    container:
-      image: cypress/browsers:node18.12.0-chrome107
-      options: --user 1001
```
gmrabian, line 178: 🤔

gmrabian (approved):

.github/workflows/deploy.yml
berryd: This breaks stuff.  It's not a real user, so it has no real permissions; no only does it not scope the paths to the workflow, but it also prevents executing container functions.

PR #14871, opened 2024-05-22
Remove the User.isActive flag

benmartin-coforma:
### Description
This PR removes a half-implemented feature; the ability to deactivate users within the app.

Background: SEDS user data is split across internal (dynamodb) and external (IDM) systems. In non-dev environments, users must exist in IDM before they can log in to SEDS. Once they exist in SEDS, details like their state list and role can be added or changed in dynamodb; from that point forward we consider dynamo to be authoritative. The isActive flag was intended to be such a detail, but even though we added code to manage this flag, we never changed any part of the app to _read_ the flag. Rather than completing that feature, we will henceforth rely on IDM for user deletion/deactivation.

This allows us to remove a good deal of code:
1. The UI elements that controlled the flag
2. The UI elements that displayed the flag
3. The UI functions which called the API to set the flag
4. The API endpoint to set the flag
5. Lots of references to the flag

Additionally, this PR contains a migration to purge the isActive flag from all existing users. A list of all users which had isActive set to false has been collected and is in the Business Owners' hands; they will do whatever is appropriate for those users in IDM and in SEDS. Once this migration has run in all environments, there will be no lingering traces of isActive anywhere in the app.

### Related ticket(s)
CMDCT-3230

---
### How to test
Cloudfront url: https://d3k0z4kfh0z5fc.cloudfront.net/

1. Log in as an admin
2. Go to the user list
   - The isActive flag should not be in the table
3. Export the user list to excel
   - The isActive flag should not be in the spreadsheet
4. Click in the user list to edit a user
   - The isActive flag should not be on the page
5. Make a change to the user and save it
   - The save should be successful

(After merging into master)
1. Confirm that the master-auth-user table has isActive flags on the users
2. Run the migration
3. Confirm that the isActive flags are now gone

### Notes
n/a

---
### Pre-review checklist

- ~[ ] I have manually tested this PR in the deployed cloud environment~
   - (except for the migration, which I tested locally and will test against dev)

---
#### Review

#### Security


gmrabian (approved):
validated admin UI shows no signs of this attribute

BearHanded (approved):

benmartin-coforma:
This migration has run in all three environments (master, val, production). The isActive flag is no more.

PR #14889, opened 2024-06-07
Fix a whole bunch of console warnings/errors

benmartin-coforma:
### Description
I looked for errors
1. In the terminal when spinning up the app locally
2. In the terminal when running Jest tests locally
3. In the browser when clicking around
4. In the github actions output when deploying

Of the 16 errors I address in this PR, 15 had no user impact. The outlier was in FormTemplates.jsx; the warning was React telling us that our labels weren't being properly attached to our inputs with the `for` attribute. Thus, this PR very slightly improved SEDS' accessibility.

I found a whole lot more errors that I am going to call out-of-scope for now. Most will require further investigation, and possibly broader changes, to fix. My notes (a stream-of-thoughts working document) are in a comment below.

### Related ticket(s)
CMDCT-3213

---
### How to test
Not much has changed, in a practical sense. Many of the most prominent warnings & errors are still present. I'll estimate that `jest test` has 50% less console output now, but the remaining 50% is still a ton.

So, idk. Make sure the unit tests pass on this PR. Maybe log in to the FormTemplates page and ensure the labels have a working `for` attributes.

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/stream-functions/package.json
+++ b/services/stream-functions/package.json
@@ -1,6 +1,7 @@
 {
   "name": "stream-functions",
   "description": "",
+  "license": "CC0-1.0",
   "resolutions": {
     "**/**/lodash": "^4.17.12",
     "minimist": "^1.2.5"
--- a/services/ui-src/src/components/FormHeader/FormHeader.jsx
+++ b/services/ui-src/src/components/FormHeader/FormHeader.jsx
@@ -42,7 +42,6 @@ const FormHeader = ({ quarter, form, year, state, updateFPL, saveForm }) => {
     const formsWithOutFPL = ["GRE"];
     async function fetchData() {
       const data = await getFormTypes();
-      console.log(data);
       const formDetails = data.find(element => element.form === form);
       setFormDescription(formDetails);
 
@@ -50,7 +49,6 @@ const FormHeader = ({ quarter, form, year, state, updateFPL, saveForm }) => {
       if (!formsWithOutFPL.includes(form)) {
         // Get answers for this form from DB
         const { answers } = await getSingleForm(state, year, quarter, form);
-        console.log(answers);
 
         // Determine Maximum FPL
         const maxFPL = getMaxFPL(answers);
--- a/services/ui-src/src/components/FormHeader/FormHeader.jsx
+++ b/services/ui-src/src/components/FormHeader/FormHeader.jsx
@@ -81,7 +81,7 @@ const FormHeader = ({ quarter, form, year, state, updateFPL, saveForm }) => {
 
   return (
     <>
-      <div row className="form-header upper-form-nav">
+      <div className="form-header upper-form-nav">
         <div className="breadcrumbs">
           <Link to="/">
             {" "}
--- a/services/ui-src/src/components/FormHeader/FormHeader.test.jsx
+++ b/services/ui-src/src/components/FormHeader/FormHeader.test.jsx
@@ -6,6 +6,8 @@ import fullStoreMock from "../../provider-mocks/fullStoreMock";
 import currentFormMock_GRE from "../../provider-mocks/currentFormMock_GRE";
 import { storeFactory } from "../../provider-mocks/testUtils";
 
+jest.spyOn(window, "alert").mockImplementation();
+
 // The props this component requires in order to render
 const defaultProps = {
   quarter: "1",
--- a/services/ui-src/src/components/FormPage/FormPage.test.jsx
+++ b/services/ui-src/src/components/FormPage/FormPage.test.jsx
@@ -7,7 +7,7 @@ import FormHeader from "../FormHeader/FormHeader";
 
 // The props this component requires in order to render
 const defaultProps = {
-  statusData: { last_modified: "01-15-2021" },
+  statusData: { last_modified: "2021-01-15T12:46:35.838Z" },
   getForm: function () {
     return;
   }
--- a/services/ui-src/src/components/FormTemplates/FormTemplates.jsx
+++ b/services/ui-src/src/components/FormTemplates/FormTemplates.jsx
@@ -106,7 +106,7 @@ const FormTemplates = () => {
         ) : null}
         {formYears && selectedYear ? (
           <div className="year-selection-container margin-top-3">
-            <label for="year-select" className="margin-top-3">
+            <label htmlFor="year-select" className="margin-top-3">
               Select Year or Create New
             </label>
             <Dropdown
@@ -122,7 +122,7 @@ const FormTemplates = () => {
         ) : null}
         {showYearInput ? (
           <>
-            <label for="year-input" className="margin-top-3">
+            <label htmlFor="year-input" className="margin-top-3">
               Enter Year
             </label>
             <TextInput
@@ -135,7 +135,7 @@ const FormTemplates = () => {
           </>
         ) : null}
         <div className="template-input margin-top-3">
-          <label for="templateInput">Enter or Modify Template</label>
+          <label htmlFor="templateInput">Enter or Modify Template</label>
           <Textarea
             id="templateInput"
             name="templateInput"
--- a/services/ui-src/src/components/GREGridWithTotals/GREGridWithTotals.jsx
+++ b/services/ui-src/src/components/GREGridWithTotals/GREGridWithTotals.jsx
@@ -299,7 +299,7 @@ const GREGridWithTotals = props => {
     return true;
   });
 
-  const totalsRow = Array.from(Array(headerCols.length - 1), (e, i) => {
+  const totalsRow = Array.from(Array(headerCols.length - 1), (_, i) => {
     let column;
 
     if (i === 0) {
@@ -310,7 +310,7 @@ const GREGridWithTotals = props => {
       );
     } else if (i === 3) {
       column = (
-        <td className="total-column">
+        <td className="total-column" key={i}>
           {addCommas(
             parseFloat(
               gridCHIPTotals.reduce(function (s, v) {
@@ -322,7 +322,7 @@ const GREGridWithTotals = props => {
       );
     } else {
       column = (
-        <td className="total-column">
+        <td className="total-column" key={i}>
           {addCommas(parseFloat(gridColumnTotals[i]).toFixed(currentPrecision))}
         </td>
       );
--- a/services/ui-src/src/components/HomeState/HomeState.test.jsx
+++ b/services/ui-src/src/components/HomeState/HomeState.test.jsx
@@ -9,11 +9,15 @@ const mockUser = {
       status: "success",
       email: "email@email.com",
       name: "Test User",
-      state: "CA",
+      states: ["CA"],
       role: "state"
     }
   ]
 };
+
+jest.mock("../../libs/api", () => ({
+  obtainAvailableForms: jest.fn().mockResolvedValue([]),
+}));
 
 const mockPush = jest.fn();
 jest.mock("react-router", () => ({
@@ -24,10 +28,7 @@ jest.mock("react-router", () => ({
 jest.mock("../../utility-functions/userFunctions", () => ({
   getUserInfo: () => Promise.resolve(mockUser)
 }));
-jest.mock("react-router", () => ({
-  ...jest.requireActual("react-router"),
-  useHistory: () => ({ push: mockPush })
-}));
+
 describe("Test HomeState.js", () => {
   let wrapper;
 
--- a/services/ui-src/src/components/Login/Login.jsx
+++ b/services/ui-src/src/components/Login/Login.jsx
@@ -118,9 +118,7 @@ export default function Login() {
         </FormGroup>
         <div className="padding-y-9">
           <LoaderButton
-            block
             type="submit"
-            bsSize="large"
             isLoading={isLoading}
             disabled={!validateForm()}
           >
--- a/services/ui-src/src/components/Print/PrintPDF.jsx
+++ b/services/ui-src/src/components/Print/PrintPDF.jsx
@@ -104,7 +104,7 @@ const PrintPDF = ({
           <h2 className="form-name">{`Form ${form} | ${formattedStateName} | ${year} | Quarter ${quarter}`}</h2>
 
           <div id="TheDiv" className="tab-container-main padding-x-5 testClass">
-            {currentTabs.map((tab, idx) => {
+            {currentTabs.map((tab, tabIndex) => {
               // Filter out just the answer objects that belong in this tab
               const tabAnswers = answers.filter(
                 element => element.rangeId === tab
@@ -114,7 +114,7 @@ const PrintPDF = ({
                 element => tab === element.rangeId
               );
               return (
-                <>
+                <React.Fragment key={tabIndex}>
                   {ageRangeDetails ? (
                     <div className="age-range-description padding-y-2">
                       <h3>{ageRangeDetails.ageDescription}:</h3>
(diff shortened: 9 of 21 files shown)


benmartin-coforma:
Here are the notes I compiled while working through this ticket. At least some of these items will find their way into future JIRA tickets.

# SEDS - Unwanted Console Output

## From terminal when running locally
* (info) We're running serverless doctor? Why? For example, with prefix `api svls doc`
* (warning) `ui-deps` complains about missing and/or incorrect peer dependencies:
  - " > @testing-library/user-event@7.2.1" has unmet peer dependency "@testing-library/dom@>=5".
  - " > @trussworks/react-uswds@7.0.0" has unmet peer dependency "@uswds/uswds@^3.7.1".
  - "aws-amplify > @aws-amplify/core > react-native-url-polyfill@1.3.0" has unmet peer dependency "react-native@*".
  - "aws-amplify > @aws-amplify/analytics > @aws-sdk/client-firehose > @aws-sdk/middleware-retry > react-native-get-random-values@1.6.0" has unmet peer dependency "react-native@>=0.56".
  - " > eslint-config-react-app@5.2.1" has incorrect peer dependency "eslint-plugin-react-hooks@1.x || 2.x".
  - " > styled-components@5.2.1" has unmet peer dependency "react-is@>= 16.8.0".
  - " > eslint-config-prettier@7.1.0" has incorrect peer dependency "eslint@>=7.0.0".
* (warning) `ui-deps` hits an error when running install script for nice-napi
  - Resolved in master, when we removed nice-napi (which was attached to nightwatch)
* (warning) `s3`: replaced cors config for bucket "local-uploads"
  - What does this mean? Why did it need to be replaced? Can we just make it good in the first place?
* (warning) `db`: @smithy/node-http-handler:WARN socket usage at capacity=50 and 135 additional requests are enqueued. See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler config.
  - happens during seed running
  - Could do as instructed, or switch to bulk commands in init scripts
* (warning) `ui`: ui| The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
* API logs are really noisy generally. Probably tone down the SDKv3 logger. Maybe disable debug logs via env variable?
* Logging in causes 5 API requests (not clear how many are from browser vs from within API)
  - obtainUserByEmail (with the email of the user you are logging in as, of course)
  - createUser!? (with null body and path params!?)
  - obtainUserByEmail again, same params
  - obtainUserByEmail TWO MORE TIMES, seemingly in parallel
  - That's too much, man. Switching to `storage` paradigm may mitigate this?

## From jest, when testing ui-src
* (warning) received `true` for non-boolean attribute `block`, at Signup.test.jsx:22:35
  - which is surely referring to `<LoaderButton block>`
  - which passes it through to a `@trussworks/react-uswds.Button`
  - which is ultimately a `<button>`, aka HTMLButtonElement
  - which doesn't have `block`
  - and neither do any of its parents: HTMLElement, Element, Node, or EventTarget
  - ✅ Removed.
* (warning) React does not recognize the `bsSize` prop, at Signup.test.jsx:22:35
  - once again it's LoaderButton
  - Maybe bsSize meant something to react-bootstrap buttons?
  - But it means nothing to us; no visual effect from removing it.
  - ✅ Removed.
* (error) NotImplemented: HTMLCanvasElement..getContext, from App.test.jsx
  - Because App needs Routes needs Users needs exportFunctions imports jspdf
  - and jspdf tries to access a Canvas, but there isn't one, because jest runs in NodeJS, not a browser
  - Could address by installing a canvas npm package... just to get this test to pass?
  - Could address by replacing JS PDF generation, which probably sucks, with a "print view" page, a la MCR
  - TODO: investigate the PDF. If it's not accessible, and not used, that's an easier argument
* (warning) bw contains an input of type undefined with both value and defaultValue props
  - From SynthesizedGrid.test.jsx, down through GridWithTotals
    + Also GREGridWithTotals; same issue copy-pasted
  - ... Yeah it is super weird that this TextInput has both value and defaultValue
  - And they don't match?? So one of them must be overwritten?
  - Or can the `value` sometimes be undefined o_O
  - TODO: dig deeper
* (warning) validateDOMNesting: tr cannot appear as a child of table
  - Within SynthesizedGrid. Easy to find, easy to fix.
  - ✅ added tbody.
* (warning): block and bsSize again, but this time in Login.jsx. Still LoaderButton.
  - ✅ removed
* (log): Stray log statement in Quarterly.jsx
  - ✅ removed
* (log): Stray log statements in FormHeader.jsx
  - ✅ removed
* (warn): Moment deprecation warning in sortingFunctions, called from Quarterly
  - I guess a date string is not formatted as expected?
  - but test behavior is not consistent >_>
  - OOH, found it. "04-08-2021"
  - ✅ fixed all misformatted last_modified dates in test data around the place
  - TODO delete the to-delete folder, it has one misformatted date I left behind
  - TODO probably also fix some other date formats, like create_date?
* (warn) a WHOOOLE bunch of warnings that updates to components are not wrapped in act(...)
  - This is going to be a lot of changes, of the same kind
  - Probably better for a separate PR
  - Will affect only tests, not runtime (just like the canvas warning above)
  - TODO in a separate story
* (warn) React does not recognize activeKey, from Nav
  - We don't specify activeKey anywhere in our code, but our Header does include react-bootstrap's Nav
  - Which does fiddle about with activeKey a bit
  - And which behaves differently based on process.env.NODE_ENV?? wth
    + I guess just to decide whether to log things to console, eg only in non-prod
  - The version of react-bootstrap used in SEDS is 0.33.1; the latest is 2.10.2
  - Maybe we can upgrade out of this one?
  - But that means upgrading from bootstrap 3.x to bootstrap 5.x, and yikes
  - TODO
* (warn)  React does not recognize activeHref, again from Nav
  - Feels safe to assume this will be resolved by whatever fixes activeKey
* (warn) Failed prop type: questionID is required by GridWithTotals, but is undefined
  - Coming from GridWithTotals, from within TabContainer.test.jsx
  - Also GREGridWithTotals, it seems. Surely inherited from GridWithTotals via copy-paste.
  - Looks like an actual bug?? Like we're trying to pull it out of somewhere it's not
  - But the page isn't grossly malfunctioning... or is it >_>
  - TODO play around with this
  - TODO then fix it
* (warn) value prop on textarea should not be null, coming from SummaryNotes
  - Seems like it'll be an empty array by default? Which is valid
  - So it must be set to null by a page interaction?
  - Probably a default value will fix this: `value={summaryNotes ?? []}`
  - ✅ default added
* (warn) can't perform a state update on an unmounted component, coming from SummaryNotes
  - Within setUserRole, line 26
  - Probably that's within a useEffect, and we have some kind of race condition
  - Probably would be resolved however we choose to resolve all the act(...) warnings
  - TODO revisit, after that's fixed.
* (warn) each child in a list should have a unique key prop, in PrintPDF.jsx:106
  - ✅ Fixed
* (warn) received true for non-boolean attribute row, in FormHeader
  - It's just on a `<div>`, so no shenanigans possible
  - ✅ Removed
* (error) not implemented: window.alert
  - For anything that calls onError in libs/errorLib, including:
    + getUserInfo
    + Profile.onLoad
  - Because Jest is running in NodeJS, not in a browser
  - Maybe we implement a shim here? 
    + `jest.spyOn(window, "alert").mockImplementation()`
  - ✅ mocked
* (error) child should have unique key, this time from GREGridWithTotals
  - ✅ Fixed
* (log) TypeError: Cannot read properties of undefined (reading '0')
  - From loadUserData, from HomeState.jsx
  - Looks manageable; however we're mocking currentUserInfo, we need Items and states to be arrays
  - ✅ Fixed, although the data is empty, and these tests now complain about act(...)

## From browser, when clicking around the app locally
* (warn) Hash History cannot PUSH the same path
  - when clicking the home button twice
  - This isn't a real problem, and I don't want to change the app's behavior
  - sooo how do we mute this?
  - TODO
* (warn) Failed to decode downloaded font: `<URL>`
  - appears 9 times in a row
  - URLs included in warning indicate ((source-sans-pro x regular, bold) + Latin-Merriweather) x woff, woff2, ttf
  - re-fetches on page resize
  - related warning: (warn) OTS parsing error: invalid sfntVersion: 1008813135
  - network shows that we're getting back whole-ass HTML pages, not font files
    + specifically, index.html
  - hopefully this is local-only?
  - TODO
* (warn) A component is changing an uncontrolled input of type undefined to be controlled.
  - When loading the user profile page for the first time (or on page refresh)
  - in Profile-div-mw-div-gw-div-gw-div-form-FormGroup2-div-FormControl2-input
  - ✅ added defaults
* (warn) `<Route>` elements should not change from controlled to uncontrolled
  - When logged in as admin, navigating from home to Users
  - "you provided a 'location' prop initially but omitted is on a subsequent render"
  - But I don't see us providing a location prop to anything? anywhere in the app.
  - This warning shows up a lot when navigating around the app, in fact
* (warn)
  - Invalid DOM property `for`. Did you mean `htmlFor`?
  - on label, at FormTemplates.jsx:138
  - why yes I did mean htmlFor, thank you. I guess.
  - very cool accessibility footgun here.
  - ✅ fixed

## From Github, during deploy
* (warn) Even more "unmet peer dependency" issues than I saw locally
  - During Jest Setup and during deploy
* (log) A whole bunch of stringified JSON during the code climate action.
  - Can we tone this down a bit maybe?
* (warn) WarmUp: skipping warmer "default" creation. No functions to warm up.
  - Are we planning on reworking our warmups at any point? No need to address this if so.
* (warn) Function generateQuarterForms has a timeout of 900 seconds, however, it's attached to API Gateway so it's automatically limited to 30 seconds.
  - Wow, really? Have we not run into this in the past?
  - TODO ask Brendan
* (warn) package.json: No license field
  - What is our license, anyway?
  - Looks like this is coming from stream-functions, and most of our packages have CC0-1.0
  - ✅ added
* (warn) Node.js will deprecate fs.rmDir, use fs.rm instead
  - Must be coming from a dependency; we don't rmdir anywhere in our code
  - Hard to tell which service though, let alone which package
  - TODO dig deeper
* (warn) The CJS build of Vite's Node API is deprecated
  - I'm sure Brendan has a good explanation
  - https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated
  - TODO
* (warn) font paths didn't resolve at build time; they will resolve at runtime instead
  - TODO confirm that this, yknow, happens
  - Boy, lots of stuff resolves at runtime :sweat:
* (warn) Some chunks are larger than 500kB after minification
  - Maybe use dynamic import() to code-split the application
  - Use build.rollupOptions.output.manualChunks to improve chunking
  - Adjust chunk size limit for this warning
  - Yeah maybe someday. The offending chunk is 1962kB, 555kB after gzip. I think.
* (warn) One deprecation found, run serverless doctor for more details
  - TODO maybe this is already resolved in master? check there.
  - doctor didn't find anything locally, but I'm looking at another branch's deploy output in GH, so maybe that's why

gmrabian (approved):

BearHanded (approved):

benmartin-coforma:
Update from 2026. We've migrated away from several of the sources of these errors, and we recently merged in #15384 - a sort of sequel PR to this one.

> * (info) We're running serverless doctor? Why? For example, with prefix `api svls doc`

✅  We no longer run serverless anything, now that we're using CDK.

> * (warning) `ui-deps` complains about missing and/or incorrect peer dependencies:

✅  `ui-deps` no longer complains about anything, because it no longer exists.

🔲 ...Although there are still some unmet peer dependency warnings, for various services. For some, we have JIRA tickets written & waiting for refinement. For others, there is nothing we can do (for example, transient dependencies which were written before React 19 (which we use now), but turn out to be perfectly compatible with it).

> * (warning) `s3`: replaced cors config for bucket "local-uploads"

✅ We no longer have a "local-uploads" bucket. Whatever this warning meant, it's gone now.

> * (warning) `db`: @smithy/node-http-handler:WARN socket usage at capacity=50 and 135 additional requests are enqueued. See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler config.

✅ Fixed by #15385

> * (warning) `ui`: ui| The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

🔲 This is still happening, actually. I suppose we'll have to look into it soon.

> * API logs are really noisy generally. Probably tone down the SDKv3 logger. Maybe disable debug logs via env variable?

🔲 Still happening. Obviously not disastrous, but maybe ought to be addressed.

> * Logging in causes 5 API requests (not clear how many are from browser vs from within API)

✅ No longer! Logging in now causes 1 API request, because the "if we can't find a user record, create one" logic has been moved to the backend.

> * (error) NotImplemented: HTMLCanvasElement..getContext, from App.test.jsx

✅ We no longer use a `canvas`, because we no longer depend on `jspdf`

> * (warning) bw contains an input of type undefined with both value and defaultValue props

✅ Fixed. I noted in 2024 that the value and defaultValue don't match, but I don't know why I thought that; they do. Rather, they did: The defaultValue has been removed.

>   * TODO delete the to-delete folder, it has one misformatted date I left behind

✅ Done.

>   * TODO probably also fix some other date formats, like create_date?

🔲 We do still have some `MM-dd-yyyy` dates in our frontend test data. Obviously a low priority.

> * (warn) a WHOOOLE bunch of warnings that updates to components are not wrapped in act(...)

✅ I never traced these back to an exact cause, but they did all disappear. Probably during the jest -> vitest migration.

> * (warn) React does not recognize activeKey, from Nav

✅ We no longer depend on bootstrap

> * (warn) Failed prop type: questionID is required by GridWithTotals, but is undefined

☑️ I think this is fixed, but I'm not sure. We will definitively resolve it as we migrate SEDS UI from JS -> TS.

> * (warn) can't perform a state update on an unmounted component, coming from SummaryNotes

☑️ I think this is fixed, although the SummaryNotes component could definitely use some simplification.

> * (warn) Hash History cannot PUSH the same path

✅ We no longer use history, because we are now on `react-router` v7.

> * (warn) Failed to decode downloaded font: `<URL>`

☑️ I never did anything to resolve this specifically, but I don't see it anymore.

> * (warn) `<Route>` elements should not change from controlled to uncontrolled

✅ I never got to the bottom of this, but it did disappear. Probably during the react router upgrade.

> * (log) A whole bunch of stringified JSON during the code climate action.

🔲 I haven't looked at this lately. Possible it's still there, but I am unbothered by it at this time.

> * (warn) WarmUp: skipping warmer "default" creation. No functions to warm up.

✅ I believe we disabled all warmups during our CDK migration.

> * (warn) Function generateQuarterForms has a timeout of 900 seconds, however, it's attached to API Gateway so 
it's automatically limited to 30 seconds.

✅ This is intentional. If the job takes more than 30 seconds (which it may), the lambda will continue running even after the gateway times out. 900 seconds should still be plenty.

> * (warn) Node.js will deprecate fs.rmDir, use fs.rm instead

☑️ I'm willing to bet this has been fixed; we're on Node 22 now. No intentional code change required on our part.

> * (warn) font paths didn't resolve at build time; they will resolve at runtime instead

🔲 For all I know, this is still happening.

> * (warn) Some chunks are larger than 500kB after minification

🔲 I don't know if this is still happening, but if so it's still in the "maybe someday" category.

PR #14899, opened 2024-06-20
[MDCT-3745] Remove unnecessary logout call

BearHanded:
### Description

Users get stuck in a loop if their token is expired when using the Okta login.

Reproducing the issue:
- SEDS refresh tokens are good for 30 days (questionable, should probably change to be in line with the other apps).
- To reproduce this in Dev I currently have it set for 1hr. Returning to that page after 1 hour reproduces the issue.
-I have only been able to reproduce it with Federated logins, because Cognito test user logins is a different auth flow.
We can't use EUA logins in dev branches

The cause:
I believe amplify 4 -> 5 causes the signout call to always navigate to the signout URL where that did not used to be the case (potentially?). When the most recent login with a federated login, the call exhibits this behavior, but not otherwise or on a user that has never logged in (incognito works around this bug).

The solution:
Remove the "convenient" signout calls done on every load of the login page. This clears nothing in the current setup but if the user had used EUA, they get stuck in a loop.

The user store already correctly identifies whether a user is active or not and shows them this page, no further signOut is necessary.

TL;DR:
Remove unnecessary logout call

### Related ticket(s)

CMDCT-3745

---
### How to test

Test that you can log in, and log out as a cognito user, but we can't test EUA until we're in dev.
https://d1dz73cleazm4z.cloudfront.net/

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/Login/Login.jsx
+++ b/services/ui-src/src/components/Login/Login.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 import { Auth } from "aws-amplify";
 import { FormGroup, FormControl, ControlLabel } from "react-bootstrap";
 import LoaderButton from "../LoaderButton/LoaderButton";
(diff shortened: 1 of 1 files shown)


gmrabian (approved):
> SEDS refresh tokens are good for 30 days 

😰

benmartin-coforma (approved):

PR #14911, opened 2024-07-11
Increase Cognito timeout to alleviate "socket hang up" issues

benmartin-coforma:
### Description
We have been seeing, across multiple environments (including prod) intermittent issues when navigating & performing actions around the site. The relevant error message seems to be this:

```
Error: Failed to fetch https://(...)/.well-known/jwks.json: write EPIPE
    at ClientRequest.oe (.../services/app-api/node_modules/aws-jwt-verify/dist/esm/https-node.js:51:25
```
Although we have also seen "socket hang up" instead of "write EPIPE" for the reason, both mean roughly the same thing: the connection was closed from the other end.

Since this issue is intermittent, my best guess is a timeout issue. I don't like that as an explanation, since this should be an AWS lambda in us-east-1, talking to an official AWS service in us-east-1, which _ought_ to be very fast. Nevertheless, if increasing the timeout for `aws-jwt-verify` addresses this issue... well, we'll have addressed the issue.

This code was cribbed from [the relevant package's readme](https://github.com/awslabs/aws-jwt-verify#configuring-the-jwks-response-timeout-and-other-http-options-with-jsonfetcher)

### Related ticket(s)
CMDCT-3826

---
### How to test
1. Deploy to an AWS environment
2. Confirm that users can still log in & do authorized actions
3. See if we ever get this error, which was intermittent to begin with

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/app-api/libs/authorization.js
+++ b/services/app-api/libs/authorization.js
@@ -32,11 +32,25 @@ export async function verifyEventSignature(event) {
 
   const { userPoolId, clientId } = await getCognitoValues();
 
-  const verifier = CognitoJwtVerifier.create({
-    tokenUse: "id",
-    userPoolId,
-    clientId,
-  });
+  const verifier = CognitoJwtVerifier.create(
+    {
+      tokenUse: "id",
+      userPoolId,
+      clientId,
+    },
+    {
+      jwksCache: new SimpleJwksCache({
+        fetcher: new SimpleJsonFetcher({
+          defaultRequestOptions: {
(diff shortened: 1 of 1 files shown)


ailZhou (approved):

BearHanded (approved):
tested this in the deployed branch, but the only way to really test it there is by repeatedly mashing the Home button. obtainUserByEmail is called on each refresh. I was unable to see an error.

Once this hits dev I will continue to stress test it.

PR #14920, opened 2024-07-25
Set the Not Applicable toggle to the correct position on page load

benmartin-coforma:
### Description
This PR addresses a cosmetic glitch that caused the "Not Applicable" toggle on every form page to be switched on, regardless of the form data's actual state.

That is: even when the form was Applicable (and the "Active" label on the left side was correctly bolded), the slider would show on the right ("Not Applicable") side on initial page load.

### Related ticket(s)
CMDCT-

---
### How to test
1. Log in as a state user
2. Navigate to any editable form (that is, a form that has not already been certified)
   * Note that the "Does this form apply to your state" toggle is on the left side, and **Active** is bolded
3. Click the toggle, and confirm that you want to lose all the data on the page
   * Note that the toggle is on the right side, and **Not Applicable** is bolded
4. Refresh the page
   * Note that the toggle and bolding have not changed
5. Feel free to keep switching the toggle and refreshing the page. You should never be able to get the toggle out of sync with the data state.

If you ever see this, then the fix isn't working:
image

### Notes

---
### Pre-review checklist

   * No, but I will in an upcoming PR!

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/NotApplicable/NotApplicable.jsx
+++ b/services/ui-src/src/components/NotApplicable/NotApplicable.jsx
@@ -21,7 +21,7 @@ const NotApplicable = ({
   saveForm,
   updatedApplicableThunk
 }) => {
-  const [applicableStatus, setApplicableStatus] = useState(1); // 0 or 1
+  const [applicableStatus, setApplicableStatus] = useState(notApplicable ? 1 : 0);
   const [disableSlider, setDisableSlider] = useState(); // Should the slider be disabled?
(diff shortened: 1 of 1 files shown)


angelaco11 (approved):
Pulled down your branch locally and went through your test steps. Everything looks good to me and the logic seems sound!

ailZhou (approved):
toggle works. not entirely sure what its overall purpose is, but it works

PR #14926, opened 2024-07-30
CMDCT-3252: fixing issue with user setting state after admin unassigning state

angelaco11:
### Description
There was a seds bug where if admin unassigned a user from a state (usually just by assigning the user to another role), upon next login of said user, there would be a 500 forbidden error when trying to select a state. This was due to a `"null"` string being passed into the state value instead of just good ol' `null`. 

All I did here was change from a "null" string to the actual null value and it resolved the issue. 

### Related ticket(s)
CMDCT-3252

---
### How to test
1. Pull down my `cmdct-3252` branch locally in seds and run locally
2. Log into seds as `adminuser@test.com`
3. Go to view/edit users
4. select a user that you have the credentials to log in with (look for the `stateuser@test.com` user) 
5. change the role of the user and make sure that when you Update User, there is no state selected
6. After the user has been logged in, log in as the user you just changed the role for
7. You should be prompted to pick a state. Select a state and make sure that the selection goes through and you are taken to another page where you can select years and quarters and so on.

### DIFF
--- a/services/ui-src/src/components/EditUser/EditUser.jsx
+++ b/services/ui-src/src/components/EditUser/EditUser.jsx
@@ -94,7 +94,7 @@ const EditUser = ({ stateList }) => {
       states = option.join("-");
     }
     if (!states) {
-      states = "null";
+      states = null;
     }
     setStatesToSend(states);
   };
@@ -122,7 +122,7 @@ const EditUser = ({ stateList }) => {
         setStatesToSend(newStates);
       } else {
         if (!e.value) {
-          e.value = "null";
+          e.value = null;
         }
         setStatesToSend([e.value]);
         response = e.value;
@@ -132,7 +132,7 @@ const EditUser = ({ stateList }) => {
     } else if (field === "role") {
       // Save to local state
       setRole(e.value);
-      setStatesToSend("null");
+      setStatesToSend(null);
       setSelectedStates();
       // Update user
       response = e.value;


benmartin-coforma (approved):
Makes sense to me!

At some point I will also want to remove [the attempt to account for this in StateSelector](https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/master/services/ui-src/src/components/StateSelector/StateSelector.jsx#L66C30-L67C29), but first I'll want to verify we don't have any users floating around with `"null"`, so that we can be sure nothing breaks.

BearHanded (commented):
In reference to Ben, there are 7 users (admin and business) in prod that have "null" as their states. This changes how we save that info, does the state selector need to read both `null` and `"null"`

benmartin-coforma:
1. I don't think it matters what the state selector does for existing users, to be honest. Because our existing users don't generally ever _see_ the state selector page. The check I linked to is internal to the state select page, so you already have to be there to be affected.
2. This change shouldn't affect the state selector anyway, since in addition to the `"null"` check I linked above, it also checks for truthiness of `user.states`.
3. I think that `string[] | undefined | null | "null"` is a terrible type, but `string[] | undefined | null` is a better one, and `string[]` would be ideal - since the lack of states can be represented by an empty array.
4. I don't fully understand the logic that lands users on the state selector page to begin with, but I would love to see a follow-up story that changes that logic, and the logic Angela modified in this PR, and the check within the state selector, and _all existing users in the database_, to simplify the type to just `string[]`.

BearHanded (approved):

PR #14937, opened 2024-08-13
Remove file upload capabilities from SEDS

benmartin-coforma:
### Description
SEDS does not now, and has not ever, allowed users to upload files. However, there was a stage of development when we thought it _might_, and code was written to support that potential future. We are now convinced that uploads will _never_ be a part of SEDS, so it is time to simplify the app by removing that code.

I have confirmed that no files exist in the production attachments bucket, so it is safe to delete.

I have confirmed that no entry of the `form-questions` table in production has `type: "file_upload"`; nor does any entry of the `form-templates` table contain the string `file_upload` at all. Therefore the schema change (`src/database/schema/form_questions.schema.json`) is definitely safe.

### Which files touched and why?
1. .env.tpl - these 2 environment variables relate to the deleted bucket, and will now be unused.
   * [ ] TODO: Remove these from the MDCT Secrets Vault in 1Password
2. .gitignore - this folder no longer exists and need not be ignored
3. README.md - We no longer use this plugin, so it doesn't need to be in our docs.
4. nightwatch/.gitignore - See 2
5. nightwatch/package.json - This package was never used by nightwatch; I think it was just copied from another package.json
   * I don't think we use nightwatch at all anyway, but that's a story for another time.
6. package.json - This package is not used at the root level. It is still used by some services, but those services will now pull it in themselves as needed.
7. serverless-compose.yml - The uploads service no longer exists, so we don't need to spin it up.
8. services/app-api/serverless.yml - This stack never used this plugin.
9. services/data-deployment/serverless.yml - see 8
10. services/database/serverless.yml - see 8
11. services/stream-functions/serverless.yml - see 8
12. services/ui-auth/serverless.yml - This stack never used this plugin, and it no longer needs to grant permissions on any S3 resources.
13. services/ui-src/.env - These 3 environment variables relate to the deleted bucket, and will now be unused.
14. services/ui-src/index.html - This line included a manifest file that had nothing to do with our app.
    * I don't know how useful it would be to include an accurate manifest file, but we certainly don't need an inaccurate one.
15. services/ui-src/package.json - this package is now unused.
16. services/ui-src/public/manifest.json - see 14
17. services/ui-src/public/scan-in-progress.html - This is a stub page appears to describes, to users, a process that does not & will not happen in SEDS.
    * I wasn't able to find any reference to this page, so I don't believe it was ever used anyway.
18. services/ui-src/public/site.webmanifest - See 14.
    * This file at least doesn't contain any _inaccurate_ information, so maybe it should be allowed to stay?
    * But it doesn't contain anything useful, either.
    * I will note that the referenced image is _not_ our site's favicon, which is defined & included elsewhere.
    * I will also note that this file was not referenced anywhere, so it seems to already be dead code.
    * [ ] TODO: I should probably either put this back, or also remove the referenced image file. Is there any way, today, for a user to see this image?
19. services/ui-src/serverless.yml - These plugins will now be unused. So will their related config and variables.
20. services/ui-src/src/config/config.js - These configuration variables will now be unused.
21. services/ui-src/src/index.jsx - These amplify settings are, once again, related to now-unused functionality.
22. services/ui-src/libs/awsLib.js - These utility functions are not referenced anywhere, and will now be unused.
23. services/ui-src/yarn.lock - Just lockfile things.
24. services/ui-waf-log-assoc/package.json - This service already depended on this plugin, but was previously relying on the root project to import it; see 6
25. services/ui-waf-log-assoc/yarn.lock - See 23.
26. services/ui-waflog-s3-bucket/package.json - See 24
27. services/ui-waflog-s3-bucket/yarn.lock - See 23.
28. services/ui/package.json - See 24
29. services/ui/yarn.lock - See 23.
30. services/uploads/* - We weren't using any part of this service.
31. src/database/schema/form_questions.schema.json - This question type was never & will never be used.
32. src/run.ts - See 7
33. yarn.lock - See 23.

### Related ticket(s)
CMDCT-3619

---
### How to test
There should be no observable changes in the app; uploads weren't supported before and they still aren't.

However, various parts of the app were touched to make this change, so there are a few things worth regression testing:
1. Can the app still be run locally?
2. Does the app still successfully deploy to AWS?
3. Can state users still log in, and navigate the site? Admin users?
4. Can form data still be saved by state users? Loaded by state and admin users?

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


rocio-desantiago (approved):
The breakdown of the files was really helpful.. Thank you!

BearHanded (approved):
Very thourough. the only followup gotcha I can think of is that the cloudformation stack that was being managed by ./run.ts is now orphaned after the removal. We'll need to manually remove the uploads stack for dev/val/prod.

PR #14947, opened 2024-09-03
Remove browserify dependency

benmartin-coforma:
### Description
Browserify is used to convert nodejs-style "`require()`" imports to a syntax that browsers can understand. But:
* The string "browserify" didn't appear in our UI code (outside of package.json and yarn.lock files, and those appearances were not invocations).
* The string "require" does not appear in our UI code (outside of yarn.lock, and a comment in Question.jsx).
* We are using Vite to build / bundle the app, which should be able to handle this type of import syntax conversion (if we ever need it (which we won't)).

Therefore, I am confident that removing the dependency from package.json will not break the app.

### Related ticket(s)
CMDCT-3850

---
### How to test
There's really nothing to test here. I guess once we merge main we should double-check that the app still comes up in the dev environment.

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/package.json
+++ b/services/ui-src/package.json
@@ -23,7 +23,6 @@
     "babel-jest": "^27.5.1",
     "babel-loader": "8.1.0",
     "babel-plugin-named-asset-import": "^0.3.6",
-    "browserify": "^17.0.0",
     "camelcase": "^5.3.1",
     "dotenv": "8.2.0",
     "dotenv-expand": "5.1.0",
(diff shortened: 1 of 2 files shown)


rocio-desantiago (dismissed):
🤔 As long it doesn't have dependencies, it should be safe to remove 👍

benmartin-coforma:
> 🤔 As long it doesn't have dependencies, it should be safe to remove 👍

Yep! None of our other packages depend on browserify. And, because of all the packages it depended on, removing browserify trims our node_modules by about 2%.

We've got a long way to go before we reach the ideal bundle size of 0kb, but the progress feels nice 😅

ailZhou (dismissed):

ailZhou (approved):

PR #14949, opened 2024-09-04
Change tealium domain for DEV ONLY

benmartin-coforma:
### Description
Per CMS requirements, we perform analytics with the 3rd-party service Tealium. Currently we load the Tealium script from a CDN, but CMS has decided to self-host the script instead. To make sure there are no surprises, they are phasing in this change one environment at a time.

This PR implements that change. In DEV only, we will begin loading Tealium from CMS' own servers.

Note that we actually load two Tealium scripts: one tiny synchronous script (utag.sync.js), and the main asynchronous one (utag.js). Previously our index.html page included three separate hardcoded script tags, to configure & load these two Tealium scripts dynamically. That's silly. I've condensed all of the Tealium-loading logic into one tag.

### Related ticket(s)
CMDCT-3955

---
### How to test
There is very little we can do to test this on our end. What we can do is verify that both <script> tags appear in the page markup when the app runs.

image

Once we deploy to val and prod, we should also verify that the script domain is correct in those environments - it should still be `https://tags.tiqcdn.com` there.

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/index.html
+++ b/services/ui-src/index.html
@@ -27,18 +27,19 @@
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" />
   <!-- Tealium Analytics Tag Manager -->
   <script>
-    var tealiumEnvMap = {
+    window.utag_cfg_ovrd = { noview: true };
+    const tealiumEnvMap = {
       "mdctseds.cms.gov": "prod",
       "mdctsedsval.cms.gov": "qa"
     };
-    var tealiumEnv = tealiumEnvMap[window.location.hostname] || "dev";
-    var tealiumDomainMap = {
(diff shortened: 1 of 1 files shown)


braxex (approved):

karla-vm (approved):

rocio-desantiago:
Oddly enough, I'm not seeing that script locally. 
Screenshot 2024-09-05 at 12 20 43 AM

I gotta figure out why I'm getting access denied for deploy link.
Screenshot 2024-09-05 at 12 21 07 AM

benmartin-coforma:
> Oddly enough, I'm not seeing that script locally.
You'll find the scripts in the `<head>`, above the `<body>`

> I gotta figure out why I'm getting access denied for deploy link.
That is strange... I know ephemeral environments don't really work for SEDS, but normally the issues I see are just data seeding problems, not actual access errors.

PR #14955, opened 2024-09-06
Remove fast-xml-parser.

benmartin-coforma:
### Description
Parsing xml is not part of our app.

I suspect this dependency was added as a workaround for a past vulnerability in a dependency. According to our lockfile today, it is only holding us back.

### Related ticket(s)
n/a

---
### How to test
n/a

### Notes
n/a

---
### Pre-review checklist

- ~[ ] I have performed a self-review of my code~
- ~[ ] I have manually tested this PR in the deployed cloud environment~

---
#### Review

#### Security

### DIFF
--- a/services/ui-auth/yarn.lock
+++ b/services/ui-auth/yarn.lock
@@ -860,13 +860,6 @@ fast-xml-parser@4.4.1:
   dependencies:
     strnum "^1.0.5"
 
-fast-xml-parser@^4.3.6:
-  version "4.4.0"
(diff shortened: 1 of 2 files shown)


services/ui-auth/yarn.lock
benmartin-coforma, line 1: Expand this diff upward to see that version 4.4.1 was and still is included as a transient dependency, with no other effects on the dependency tree.

BearHanded (approved):

PR #14970, opened 2024-10-28
IaC change - Replace Serverless Framework with AWS CDK

JonHolman:
### Description
This is our first pull request to move away from the serverless framework and replace it with AWS CDK.  This needs extensive peer review.  We still need to figure out:
1. Data migration
2. Running Locally.

### Related ticket(s)

CMDCT-

---
### How to test

1. This branch is deployed to https://d2k762g1ngqwda.cloudfront.net/
2. You can also create your own branch and github actions will deploy as expected.

### Notes

---
### Pre-review checklist

- [X] I have performed a self-review of my code
- [X] I have manually tested this PR in the deployed cloud environment

---
#### Review

#### Security

---


lib/config/deployment-config.ts
```diff
@@ -0,0 +1,113 @@
...
+  iamPath: string;
+  iamPermissionsBoundary: string;
+  vpcName: string;
+};
+
+export type DeploymentConfigProperties = InjectedConfigProperties & {
+  isDev: boolean;
+  project: string;
+  sharedOpenSearchDomainArn: string;
+  sharedOpenSearchDomainEndpoint: string;
```
BearHanded, line 22: I think these openSearch ones are leftover from Mako

lib/config/deployment-config.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  public static async fetch(
+    options: InjectedConfigOptions
+  ): Promise<DeploymentConfig> {
+    const injectedConfig = await DeploymentConfig.loadConfig(options);
+    const appConfig: DeploymentConfigProperties = {
+      ...injectedConfig,
+      project: options.project,
+      stage: options.stage,
+      isDev: !["main", "val", "production"].includes(options.stage),
```
BearHanded, line 42: seds currently uses master/val/production. If we're swapping to main this lines up

lib/config/deployment-config.ts
```diff
@@ -0,0 +1,113 @@
...
+      terminationProtection: ["main", "val", "production"].includes(
```
BearHanded, line 43: ditto

lib/config/deployment-config.ts
```diff
@@ -0,0 +1,113 @@
...
+        options.stage
+      ),
+      sharedOpenSearchDomainArn: "",
```
BearHanded, line 46: more Mako endpoints

lib/local-constructs/dynamodb-table/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  public readonly table: dynamodb.Table;
+
+  constructor(scope: Construct, id: string, props: DynamoDBTableProps) {
+    super(scope, id);
+
+    this.table = new dynamodb.Table(this, "Table", {
+      tableName: `${props.stage}-${props.name}`,
+      partitionKey: props.partitionKey,
+      billingMode: dynamodb.BillingMode.PAY_PER_REQUEST,
+      stream: dynamodb.StreamViewType.NEW_AND_OLD_IMAGES,
```
BearHanded, line 25: I don't think all our tables need a stream, but it definitely doesn't hurt? If it's not that way in SEDS it definitely has some in other apps that don't have streams

lib/local-constructs/lambda/index.ts
```diff
@@ -0,0 +1,186 @@
...
+        const currentTable = Stack.of(table)
+          .getLogicalId(table.node.defaultChild as CfnElement)
+          .slice(0, -8);
+
+        acc[`${currentTable}Name`] = table.tableName;
+        acc[`${currentTable}Arn`] = table.tableArn;
+
+        return acc;
+      }, {} as { [key: string]: string }),
+    };
```
BearHanded, line 69: Are there any env vars for lambdas that aren't tables? I'll have to look but I would guess there's at least one. Maybe the pdf gen one.

lib/local-constructs/waf/index.ts
BearHanded, line 1: Has this construct been setup to optionally limit to the VPN? Our current WAF setup has that capability and it is enabled for QMR dev

src/write-ui-env-file.ts
```diff
@@ -0,0 +1,55 @@
...
+    SKIP_PREFLIGHT_CHECK: "true",
+    API_REGION: region,
+    API_URL: deploymentOutput.apiGatewayRestApiUrl.slice(0, -1),
+    COGNITO_REGION: region,
+    COGNITO_IDENTITY_POOL_ID: deploymentOutput.identityPoolId,
+    COGNITO_USER_POOL_ID: deploymentOutput.userPoolId,
+    COGNITO_USER_POOL_CLIENT_ID: deploymentOutput.userPoolClientId,
+    COGNITO_USER_POOL_CLIENT_DOMAIN: deploymentOutput.userPoolClientDomain,
+    COGNITO_REDIRECT_SIGNIN: local
+      ? "http://localhost:5000/"
```
BearHanded, line 33: placeholder port?

lib/stacks/api.ts
```diff
@@ -0,0 +1,437 @@
...
+    };
+
+    new Lambda(this, "ForceKafkaSync", {
+      entry: "services/app-api/handlers/kafka/get/forceKafkaSync.js",
+      timeout: cdk.Duration.minutes(15),
+      memorySize: 3072,
+      ...commonProps,
+    });
+
+    const postKafkaData = new Lambda(this, "postKafkaData", {
```
BearHanded, line 121: We should check this is still running successfully when we're editing data in the app.

The logs are extremely chatty when it can't connect or something goes wrong, so just vetting that postKafkaData is returning successes.

lib/local-constructs/waf/index.ts
JonHolman: Not yet.  I saw the boolean in the serverless.yml but they were all set to false, so I did not add that functionality yet.

lib/config/deployment-config.ts
JonHolman: Thank you

lib/stacks/data.ts
```diff
@@ -0,0 +1,156 @@
...
+  public readonly tables: { [name: string]: dynamodb.Table };
+  public readonly seedDataFunction: lambda_nodejs.NodejsFunction | undefined;
+
+  constructor(scope: Construct, id: string, props: DatabaseStackProps) {
+    super(scope, id, props);
+
+    const stage = this.node.tryGetContext("stage") || "dev";
+
+    this.tables = {
+      "age-ranges": new DynamoDBTable(this, "AgeRanges", {
```
BearHanded, line 28: these are the same tables defined in data, yea? should this be a shared definition somewhere? or can api consume this?

BearHanded: wait is this just a duplicate of the data stack? where is the api definitions?

cdk.json
```diff
@@ -0,0 +1,70 @@
...
+    "exclude": [
+      "README.md",
+      "cdk*.json",
+      "**/*.d.ts",
+      "**/*.js",
+      "tsconfig.json",
+      "package*.json",
+      "node_modules",
+      "test",
+      "bun.lockb"
```
gmrabian, line 16: we don't use bun so I think we can remove this line

cdk.json
```diff
@@ -0,0 +1,70 @@
...
+    ]
+  },
+  "context": {
```
gmrabian, line 19: i'd appreciate a rundown or a link to some documentation on all these choices. I see ECS and EC2 in there which we don't use so just want to learn more about what they are doing

gmrabian (commented):
initial thoughts. still reviewing

cdk.json
JonHolman: Good question and I agree.  Unfortunately the cdk.json file is not very well documented, that file is very close to what you get for a cdk.json when you run:
```
cdk init app --language typescript
```
in a new directory to start a new project.  Throughout CDK's documentation they make references to change certain behaviors, change this or that in the cdk.json, I always start with the defaults the init provides.  I agree about EC2 and EC2, but I haven't tried removing them as they're there by default.  We could try, the CDK tooling may panic.

cdk.json
JonHolman: Thanks

lib/local-constructs/dynamodb-table/index.ts
JonHolman: Yeah, the postkafka lambda connects to every dynamodb table's stream.  dataconnect's doesn't use every table's stream, but postkafka has all of them.

lib/local-constructs/lambda/index.ts
JonHolman: There's the broker string and stage.

src/run.ts
```diff
@@ -162,35 +210,143 @@ async function deploy(options: { stage: string }) {
...
-    },
-  ];
-  if (options.service) {
-    filters.push({
-      Key: "SERVICE",
-      Value: `${options.service}`,
-    });
+  const stackName = `${project}-${stage}`;
+
+  if (/prod/i.test(stage)) {
```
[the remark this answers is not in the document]
JonHolman: would that just be "prod" in all lowercase or should we do something like:
```
["prod", "production"].includes(stage.toLowerCase()
```
are we worried about additional permutations?

PR #14988, opened 2024-12-03
Improvements to CDK project - frontend deployment and various others

JonHolman:
### Description
This branch started with updating the CDK to incorporate the react frontend build and deployment into the CDK deployment rather than being accompanied via a wrapper script that ran with the deployment.  But... then due to the holidays and not wanting to create a PR during that time... I kept working on the same branch so it grew and it also has changes to:

1. Stop getting SSM parameters throughout the deployment process and just get them all at the beginning, that avoids any of the deployment functions ever needing to be async (CDK doesn't like that).
2. Added a construct specifically for lambda functions that are to subscribe to dynamodb event streams.
3. Leverage AWS CDK custom resources to invoke our bootstrap users and seed data functions, that's the CDK way to do it.
4. Improving the WAF rules for VPN restriction when enabled.
5. Misc cleanup.

Most, if not all of these are improvements based on Ben's feedback on my initial PR, thanks again Ben.

### Related ticket(s)

CMDCT-

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/constructs/cloudwatch-logs-resource-policy.ts
+++ b/deployment/constructs/cloudwatch-logs-resource-policy.ts
@@ -1,6 +1,6 @@
 import { Construct } from "constructs";
 import { CfnResourcePolicy } from "aws-cdk-lib/aws-logs";
-import { Stack, aws_iam as iam } from "aws-cdk-lib";
+import { Aws, aws_iam as iam } from "aws-cdk-lib";
 
 interface CloudWatchLogsResourcePolicyProps {
   readonly project: string;
@@ -15,7 +15,6 @@ export class CloudWatchLogsResourcePolicy extends Construct {
     props: CloudWatchLogsResourcePolicyProps
   ) {
     super(scope, id);
-    const stack = Stack.of(this);
 
     const policyDocument = new iam.PolicyDocument({
       statements: [
@@ -24,14 +23,14 @@ export class CloudWatchLogsResourcePolicy extends Construct {
           principals: [new iam.ServicePrincipal("delivery.logs.amazonaws.com")],
           actions: ["logs:CreateLogStream", "logs:PutLogEvents"],
           resources: [
-            `arn:aws:logs:*:${stack.account}:log-group:aws-waf-logs-*`,
-            `arn:aws:logs:*:${stack.account}:log-group:/aws/http-api/*`,
-            `arn:aws:logs:*:${stack.account}:log-group:/aws/vendedlogs/*`,
+            `arn:aws:logs:*:${Aws.ACCOUNT_ID}:log-group:aws-waf-logs-*`,
+            `arn:aws:logs:*:${Aws.ACCOUNT_ID}:log-group:/aws/http-api/*`,
+            `arn:aws:logs:*:${Aws.ACCOUNT_ID}:log-group:/aws/vendedlogs/*`,
           ],
           conditions: {
-            StringEquals: { "aws:SourceAccount": stack.account },
+            StringEquals: { "aws:SourceAccount": Aws.ACCOUNT_ID },
             ArnLike: {
-              "aws:SourceArn": `arn:aws:logs:${stack.region}:${stack.account}:*`,
+              "aws:SourceArn": `arn:aws:logs:${Aws.REGION}:${Aws.ACCOUNT_ID}:*`,
             },
           },
         }),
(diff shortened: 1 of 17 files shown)


deployment/stacks/api.ts
```diff
@@ -546,5 +522,8 @@ export function createApiComponents(props: CreateApiComponentsProps) {
     props.iamPath
   );
 
-  return { restApiId: api.restApiId, apiGatewayRestApiUrl: api.url };
+  return {
+    restApiId: api.restApiId,
+    apiGatewayRestApiUrl: api.url.slice(0, -1),
```
peoplespete, line 527: Pardon my ignorance, what does the slice part do here?

deployment/constructs/cloudwatch-to-s3.ts
```diff
@@ -1,94 +1,99 @@
...
 
   constructor(scope: Construct, id: string, props: CloudWatchToS3Props) {
     super(scope, id);
 
     const { logGroup, bucket, filePrefix = "", filterPattern = "" } = props;
 
     // Create a Firehose role
-    const firehoseRole = new Role(this, "FirehoseRole", {
-      assumedBy: new ServicePrincipal("firehose.amazonaws.com"),
+    const firehoseRole = new iam.Role(this, "FirehoseRole", {
```
peoplespete, line 28: I like this `iam.` style of using imports!

deployment/constructs/cloudwatch-logs-resource-policy.ts
```diff
@@ -24,14 +23,14 @@ export class CloudWatchLogsResourcePolicy extends Construct {
           principals: [new iam.ServicePrincipal("delivery.logs.amazonaws.com")],
           actions: ["logs:CreateLogStream", "logs:PutLogEvents"],
           resources: [
-            `arn:aws:logs:*:${stack.account}:log-group:aws-waf-logs-*`,
-            `arn:aws:logs:*:${stack.account}:log-group:/aws/http-api/*`,
-            `arn:aws:logs:*:${stack.account}:log-group:/aws/vendedlogs/*`,
+            `arn:aws:logs:*:${Aws.ACCOUNT_ID}:log-group:aws-waf-logs-*`,
```
peoplespete, line 26: Ohh this is great!

deployment/constructs/lambda-dynamo-event.ts
```diff
@@ -0,0 +1,102 @@
...
+  extends Partial<lambda_nodejs.NodejsFunctionProps> {
+  additionalPolicies?: iam.PolicyStatement[];
+  brokerString?: string;
+  handler: string;
+  iamPath: string;
+  iamPermissionsBoundary: iam.IManagedPolicy;
+  memorySize?: number;
+  stackName: string;
+  tables: dynamodb.Table[];
+  timeout?: Duration;
```
peoplespete, line 20:
```suggestion
  iamPath: string;
  iamPermissionsBoundary: iam.IManagedPolicy;
  stackName: string;
  tables: dynamodb.Table[];
```
Looks like these ones are already on [NodejsFunctionProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.NodejsFunctionProps.html)

deployment/stacks/api.ts
JonHolman: just slices off the last character which is the slash at the end of the url.  In seds I noticed the react frontend adds the / at the beginning when its creating the URL for an API endpoint, so if send that URL as is it ends up sending // which results in an error.

benmartin-coforma: We could make this a little more self-documenting with a regex that specifies we're removing a trailing slash: `api.url.replace(/\/$/, "")`

Wait no that's way worse, never mind.

It may be worthwhile in the future to revisit _why_ our frontend is blindly appending that `/`, make it not do that, and come back to simplify this code. On the other hand, that may not be worthwhile. If it ain't broke, right?

deployment/stacks/parent.ts
benmartin-coforma, line 1: This file is looking _really_ pretty now. It lays out all the major pieces that need to be created, and how they relate to each other, in a really readable way. I'm a big fan.

benmartin-coforma (approved):
Outstanding work. I love seeing that section of `run.ts` disappear, because it pales in comparison to `deployFrontend.ts`

peoplespete (approved):
🏄

PR #14990, opened 2024-12-05
CMDCT-4184 sls-side prepping serverless to retain ApiGateway when removed

peoplespete:
### Related ticket(s)

CMDCT-

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -91,149 +91,149 @@ jobs:
       BRANCH_SPECIFIC_VARNAME_AWS_DEFAULT_REGION: ${{ steps.set_names.outputs.BRANCH_SPECIFIC_VARNAME_AWS_DEFAULT_REGION }}
       BRANCH_SPECIFIC_VARNAME_AWS_OIDC_ROLE_TO_ASSUME: ${{ steps.set_names.outputs.BRANCH_SPECIFIC_VARNAME_AWS_OIDC_ROLE_TO_ASSUME }}
 
-  register-runner:
-    name: Register GitHub Runner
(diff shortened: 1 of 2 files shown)


services/ui/serverless.yml
```diff
@@ -106,16 +106,6 @@ resources:
         - Fn::Equals:
             - ""
             - ${self:custom.iamPermissionsBoundaryPolicy}
-    CreateDnsRecord:
```
peoplespete, line 109: This was never actually used.

services/ui/serverless.yml
```diff
@@ -213,60 +203,19 @@ resources:
           Comment: OAI to prevent direct public access to the bucket
     CloudFrontDistribution:
       Type: AWS::CloudFront::Distribution
+      DeletionPolicy: Retain
```
peoplespete, line 206: This allows us to keep this resource even after we've destroyed the serverless assets.

JonHolman:
All of this except for modifying the CloudFront distribution can be done immediately.  How about we do it now, so that we can get it over with?

peoplespete:
> All of this except for modifying the CloudFront distribution can be done immediately. How about we do it now, so that we can get it over with?

We decided to leave it here for now since all of the retains would need to be conditional if we merge it soon and rather than add that complexity and since we can't entirely get rid of pete-sls bc of CloudfrontDistribution changes, we'll keep it in the PR and then pete-sls.

JonHolman (approved):

benmartin-coforma (approved):

PR #14991, opened 2024-12-09
CMDCT-4184 cdk-side prepping cdk to import Cloudfront Distribution

peoplespete:
### Description
This change is not to be deployed until we are ready to convert to cdk. For full details of the order of events recommended to preserve our Cloudfront Distribution check out this readme: https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/cmdct-4184-cdk/import_instructions.md
The idea is that by preserving Cloudfront Distribution we will be able to deploy cdk without requiring any external DNS changes. 🎉 

### Related ticket
CMDCT-4184

---
### How to test
If you want to you can do the following to test this PR:
1) checkout pete-sls (matching master at the moment) branch
2) deploy serverless via run script (check that cloudfront distribution points at app front end)
3) checkout cmdct-4184-sls
4) deploy serverless via run script (check that cloudfront distribution is stripped down and no longer points at app front end)
5) destroy serverless assets via ./run destroy command
6) verify that distribution still exists
7) checkout cmdct-4184-cdk and comment out all code for the distribution (and things that depend on it)
8) deploy cdk via run command
9) comment IN the simplified distribution section
10) run cdk import
11) deploy cdk via run command
12) comment OUT the simplified distribution section and comment IN the final distribution section
12) deploy cdk via run command
13) verify that the cloudfront distribution is now in the cdk stack and points to the application.

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---


deployment/stacks/ui.ts
```diff
@@ -136,7 +148,6 @@ export function createUiComponents(props: CreateUiComponentsProps) {
   const applicationEndpointUrl = `https://${distribution.distributionDomainName}/`;
 
   setupWaf(scope, stage, project, deploymentConfigParameters);
-  setupRoute53(scope, distribution, deploymentConfigParameters);
```
peoplespete, line 139: This was never actually being created or used.

deployment/stacks/ui.ts
```diff
@@ -95,6 +93,20 @@ export function createUiComponents(props: CreateUiComponentsProps) {
     }
   );
 
+  // new cloudfront.Distribution(
```
peoplespete, line 96: This is the simplified version of the distribution that will initially be used for import. Once imported we'll update it to bring it back into the fully featured setup Jon got working for cdk.

deployment/import_instructions.md
```diff
@@ -0,0 +1,114 @@
...
+cognito.UserPool -
+```
+
+3. Destroy sls
+
+```
+./run destroy --stage <YOUR_BRANCH_NAME>
+```
+
+:warning: Make sure that all sls associated stacks have been fully destroyed before proceeding.
```
JonHolman, line 23: Is there a way we can validate that the retain setting took before we destroy the stack?

deployment/import_instructions.md
```diff
@@ -0,0 +1,114 @@
...
+```
+As this import occurs you'll have to provide the information you gathered just before destroying the serverless stacks. For the dynamo tables the default should be correct but read closely to be sure.
+
+3. Run a deploy on that same imported resource set.
+
+```bash
+WITH_IMPORTS=true ./run deploy --stage <YOUR_BRANCH_NAME>
+```
+
+4. Run a full deploy by kicking off the full cdk deploy via Github Action. Permissions for individual developers are limited so you must use Github Action to do this part.
```
JonHolman, line 48: we can't do this from our local?

deployment/import_instructions.md
```diff
@@ -0,0 +1,114 @@
...
+5. Find the Cloudfront Url in the Github Action's logs (or in the outputs section of your Cloudformation Stack). Visit the site and confirm that you can login and use the application. :tada: Congrats, you did it!
+
+
+## What if it all goes pear shaped?
+
+### If during the middle of the migration, things begin to break and we need to reinstate the serverless stack, we need a way to bring the Cloudfront Distribution back into the newly rebuilt serverless stack. Fortunately this is possible if you follow these steps.
+
+:grey_exclamation: These instructions are specific to reimporting a Cloudfront Distribution but the same pattern should also apply to any other imported resources that need un-importing should the need arise.
+
+1) Get the Cloudfront Distribution unaffiliated with any Cloudformation stack. If it's already been successfully imported into the new cdk stack then you'll need to destroy the cdk stack to eject it from that stack.
```
JonHolman, line 59: do you need to retain the resources first or are they already set to retain in the CDK stack?

deployment/import_instructions.md
peoplespete: This was done as CMDCT-4238 which you did, thanks!

deployment/import_instructions.md
peoplespete: So for Dynamo retaining is the default: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_dynamodb.Table.html#removalpolicy

For Cognito user pool retaining is the default:
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cognito.UserPool.html#removalpolicy

Another ticket added a conditional removal policy to the cdk Cloudfront Distribution already.

One option would be replace dynamo and cognito user pool's implicit assumption that AWS won't change their defaults to explicitly define retention. What do you think?

deployment/import_instructions.md
peoplespete: Just confirmed, you're right we can. I've updated the instructions accordingly.

deployment/import_instructions.md
```diff
@@ -0,0 +1,119 @@
...
+
+```
+./run deploy --stage <YOUR_BRANCH_NAME>
+```
+
+2. Collect information about the resources we're going to be importing into the new cdk stack.
+```
+cloudfront.Distribution -
+cognito.UserPool -
+```
```
JonHolman, line 15: DynamoDB tables should be here as well, right?

deployment/import_instructions.md
```diff
@@ -0,0 +1,119 @@
...
+
+3. Destroy sls
+
+```
+./run destroy --stage <YOUR_BRANCH_NAME>
```
JonHolman, line 20: We need to ensure the destroy updates to the run script I made to master are in the pete-sls branch as well.

JonHolman (approved):

deployment/app.ts
```diff
@@ -1,57 +1,62 @@
...
-        account: process.env.CDK_DEFAULT_ACCOUNT,
-        region: process.env.CDK_DEFAULT_REGION,
-      },
-      deploymentConfigParameters,
-    });
-  } catch (error) {
-    console.error("Error:", error);
-    process.exit(1);
+  let correctParentStack;
+  if (process.env.WITHOUT_IMPORTS) {
```
peoplespete, line 45: Change to one env variable that switches and we use the string value comparison to flip between stacks (and we default to ParentStack).

benmartin-coforma (approved):

PR #15004, opened 2024-12-14
Replace react-data-table-component

benmartin-coforma:
### Description
This PR replaces the `react-data-table-component` library with bare `<table>` HTML markup. We were using the library for two tables:
* The list of forms available for a given state, year, and quarter (Quarterly.jsx)
* The list of users that exist in the site, which is only viewable by admins (Users.jsx)

Why replace `react-data-table-component`?
* This will resolve PR #14992 (an automated PR to bump the version of this library), and all of its future successor PRs, permanently. The table markup we are replacing it has been valid in all browsers for decades, and we can expect it to remain valid for decades hence, with no further updates.
* There's an argument to be made that the new table markup is more accessible; I'm not sure how much water this holds, but I am very pleased that our markup is now an actual table, rather than the divs in divs in divs this library generated.
* This also removes a blocker for SEDS' eventual React upgrade. Yarn throws warnings about `incorrect peer dependencies` when I try to upgrade to React 18, because `react-data-table-component` was built for React 16. Does it still work fine under React 18? I don't know, maybe! Do you want to test it? Because I don't.
* We were not using most of the features `react-data-table-component` provides. It boasts selectable rows, expandable rows, pagination, "themability", and "responsiveness". I don't know what responsiveness means in this context, but as far as theming... The new table markup looks better with less CSS than we spent wrangling the library.
* This also removes our dependency on `react-data-table-extensions`, which I'm not convinced we were actually making use of at all.
* This also removes our dependency on `styled-components`, the existence of which personally offends me.
* Shipping less code for the tables means less bandwidth and faster load times, in theory. I can't brag too hard about this one because I didn't measure it; in all likelihood it is completely overshadowed by other factors, and wasn't my original motivation regardless.

Stylistic & behavioral changes:
* The Users table previously had built-in search, and it no longer does. If an admin is searching within this table to find a specific user, they will find their browser's built-in search to be more capable and more performant.
* Neither table supports client-side sorting; you get the order you get.
   - Sorting never made sense for the forms - there are only 5 of them.
   - In theory sorting may have been useful in the user table - perhaps sorting by last name, and scrolling to the user you're looking for? But you can also just search for that user, so this too was of little use.
* The table rows have slightly less vertical padding
* The form statuses in the Quarterly Forms table are no longer buttons. They still look kind of like buttons, but we can do further restyling to fix that in a future PR
* The tables no longer have every-other-row shading
* The table rows no longer change color on hover
* The table columns now automatically change width to accommodate the data you put in them - so form names and user emails no longer need to wrap, and columns containing only buttons or abbreviations are narrower as appropriate. This was done with less code than before! HTML really tries hard to make your page look nice, actually. You just need to stop telling it not to.

Quarterly Forms table:
image

Users table:
image

### Related ticket(s)
n/a

---
### How to test
To test the Quarterly Forms table
1. Log in as `adminuser@test.com` or `stateuser@test.com` (both roles see the same table, so it does not matter which you choose)
2. If you logged in as an admin, select a state. I usually choose Alabama; it's one of the states for which we seed data in test environments.
3. Select a year and quarter
4. You should see the table of forms. It will have about 5 rows, but that may vary by state and quarter
5. The table should look good.

To test the Users table
1. Log in as `adminuser@test.com`
2. Click "View/Edit Users"
3. You should see the user table.
4. The table should look good.
5. The search function should work.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


ailZhou (approved):
lgtm! tested locally, table is solid

karla-vm (dismissed):
This is neat! ✨ if I had the final say, I would remove the "Search" bar but I would defer to Product and/or Design on whether or not that is an option for us in SEDS. Otherwise, looks awesome!

ailZhou (dismissed):

ailZhou (approved):

karla-vm (approved):

PR #15012, opened 2024-12-19
CMDCT-4194 cdk side, creating a better way to import resources from serverless including Cognito UserPool

peoplespete:
### Description
This mostly tweaks the way we are doing the importing so that the main cdk setup can be used as is and we just run special commands during importing. For more details check out `deployment/import_instructions.md`.

### Related ticket(s)
CMDCT-4194

---
### How to test
You are welcome to try following the `deployment/import_instructions.md` and see if it all makes sense and enables you to import resources from a serverless stack.

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/stacks/data.ts
+++ b/deployment/stacks/data.ts
@@ -79,25 +79,28 @@ export function createDataComponents(props: CreateDataComponentsProps) {
         "service-role/AWSLambdaVPCAccessExecutionRole"
       ),
     ],
+    inlinePolicies: {
+      DynamoPolicy: new iam.PolicyDocument({
+        statements: [
+          new iam.PolicyStatement({
+            effect: iam.Effect.ALLOW,
+            actions: [
+              "dynamodb:DescribeTable",
+              "dynamodb:Query",
+              "dynamodb:Scan",
+              "dynamodb:GetItem",
+              "dynamodb:PutItem",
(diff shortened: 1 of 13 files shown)


deployment/stacks/data.ts
```diff
@@ -79,26 +79,29 @@ export function createDataComponents(props: CreateDataComponentsProps) {
         "service-role/AWSLambdaVPCAccessExecutionRole"
       ),
     ],
+    inlinePolicies: {
```
peoplespete, line 82: This and in one other spot I added this minor refactor to add the inline policies directly to the resources.

src/run.ts
```diff
@@ -209,7 +209,7 @@ async function deploy(options: { stage: string }) {
   const stage = options.stage;
   const runner = new LabeledProcessRunner();
   await prepare_services(runner);
-  const deployCmd = ["cdk", "deploy", "--context", `stage=${stage}`, "--all"];
+  const deployCmd = ["cdk", "deploy", "--context", `stage=${stage}`, "--method=direct", "--all"];
```
peoplespete, line 212: This should make the deploys a bit faster. Instead of using a changeset it directly updates the stack. Apparently it makes it faster, but you lose some progress tracking while it is running. I figured that was better. 🤷

PR #15015, opened 2024-12-23
CMDCT-4186 - cdk side, "import" by recreating dynamo tables so that they are retained

peoplespete:
### Description
We're going with a direct import on Dynamo tables. This is fine as long as the stage is named the same between sls and cdk stacks. That should be true for all the import stacks like master/val.

### Related ticket(s)
CMDCT-4186

---
### How to test
Follow `deployment/import_instructions.md` if you'd like to manually try this out.

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---


deployment/constructs/dynamodb-table.ts
```diff
@@ -29,7 +29,8 @@ export class DynamoDBTable extends Construct {
   constructor(scope: Construct, id: string, props: DynamoDBTableProps) {
     super(scope, id);
 
-    const tableName = `${props.stage}-${props.name}`;
+    // const tableName = `${props.stage}-${props.name}`;
+    const tableName = `cmdct-4186-sls-${props.name}`;
```
peoplespete, line 33: This is a tricky part. We have to use the old dynamo names but then keep using them always. And ideally do this in a way that will work for dev branches, though that is probably a nice to have.

deployment/constructs/dynamodb-table.ts
```diff
@@ -23,21 +23,20 @@ export interface DynamoDBTableIdentifiers {
...
-  public readonly table: dynamodb.Table;
+  public readonly table: dynamodb.TableV2;
   public readonly identifiers: DynamoDBTableIdentifiers;
 
   constructor(scope: Construct, id: string, props: DynamoDBTableProps) {
     super(scope, id);
 
     const tableName = `${props.stage}-${props.name}`;
-    this.table = new dynamodb.Table(this, "Table", {
+    this.table = new dynamodb.TableV2(this, "Table", {
```
peoplespete, line 33: Switching to v2 is recommended esp if we go this route where the tables change to a new table with a new name. Advantage according to AWS is that with V2 everything is a "Global Table" meaning you can have cross region replicas but you can't take a v1 table and do that so they recommend making all tables as global tables now.

deployment/import_instructions.md
```diff
@@ -26,25 +33,27 @@ cognito.UserPool -
 1. Create just the new cdk stack without anything inside of it.
 
 ```bash
-WITHOUT_IMPORTS=true ./run deploy --stage <YOUR_BRANCH_NAME>
+WITHOUT_IMPORTS=true FORM_ANSWERS=FILL_IN_WITH_VALUES_ABOVE FORM_QUESTIONS=FILL_IN_WITH_VALUES_ABOVE FORM_TEMPLATES=FILL_IN_WITH_VALUES_ABOVE FORMS=FILL_IN_WITH_VALUES_ABOVE STATE_FORMS=FILL_IN_WITH_VALUES_ABOVE STATES=FILL_IN_WITH_VALUES_ABOVE AUTH_USER=FILL_IN_WITH_VALUES_ABOVE ./run deploy --stage <YOUR_BRANCH_NAME>
```
peoplespete, line 36: The reason this is working this way is that if we import, we have to keep the old names which is not great because you always have to remember the old names and wouldn't match the pattern for new stacks we make with the same cdk code. So I'm taking the dynamo tables we would normally import and using point-in-time-recovery to create an identical table correctly named and managed by the cdk code. I've tried to do it such that in the future we can delete the import code someday and there will be no indication that it ever existed and there will be no weirdness from unconventionally named tables.

JonHolman:
I think we should consider making if statements in files, rather than having directories with_imports and without_imports.  Also, if we're going to do with_imports and without_imports, the original file should be git moved somewhere to preserve the history.

deployment/stacks/without_imports/parent.ts
```diff
@@ -12,5 +13,11 @@ export class WithoutImportsParentStack extends Stack {
     props: StackProps & DeploymentConfigProperties
   ) {
     super(scope, id, props);
+
+    const {
+      stage,
+    } = props;
+
+    createDataComponents(stage);
```
JonHolman, line 21: since we're only getting stage from the props, can we do props.stage and eliminate lines 17-19?

JonHolman:
Our CDK table names don't match our SLS table names?

peoplespete:
@JonHolman This is fixed up and I tested that it actually does work with a stack.
If you want to poke around the deploys and stack, I used `pete-test` as a branch to do it from start to finish.

peoplespete:
> I think we should consider making if statements in files, rather than having directories with_imports and without_imports. Also, if we're going to do with_imports and without_imports, the original file should be git moved somewhere to preserve the history.

The reason I like this strategy is that at the end of all of the migration, once we're confident that everything is looking good we can remove those directories and the stack behaves as it the imports never occurred. Other than that, we just have the one "which parent stack" if to remove. 

This makes sense to me since the best import strategy is to 1) deploy an empty stack, 2) import the sls assets, and then 3) deploy everything. The ifs to make that happen seemed to proliferate when I was trying to do it all inside of the single cdk setup. I'm approaching it this way because so many resources depend on the imported resources. It gets really complicated if you try and build most of the cdk stack before importing the resources.

peoplespete:
> Our CDK table names don't match our SLS table names?

It does now good catch!

JonHolman:
> > I think we should consider making if statements in files, rather than having directories with_imports and without_imports. Also, if we're going to do with_imports and without_imports, the original file should be git moved somewhere to preserve the history.
> 
> The reason I like this strategy is that at the end of all of the migration, once we're confident that everything is looking good we can remove those directories and the stack behaves as it the imports never occurred. Other than that, we just have the one "which parent stack" if to remove.
> 
> This makes sense to me since the best import strategy is to 1) deploy an empty stack, 2) import the sls assets, and then 3) deploy everything. The ifs to make that happen seemed to proliferate when I was trying to do it all inside of the single cdk setup. I'm approaching it this way because so many resources depend on the imported resources. It gets really complicated if you try and build most of the cdk stack before importing the resources.

How about removing the without imports directory and parent.ts file?  Since that's just calling our original, right?

JonHolman (approved):

peoplespete:
> How about removing the without imports directory and parent.ts file? Since that's just calling our original, right?

The without imports directory just has a parent file. All it does it create an empty stack. Here's what it does:

```ts
import { Construct } from "constructs";
import {
  Stack,
  StackProps,
} from "aws-cdk-lib";
import { DeploymentConfigProperties } from "../../deployment-config";

export class WithoutImportsParentStack extends Stack {
  constructor(
    scope: Construct,
    id: string,
    props: StackProps & DeploymentConfigProperties
  ) {
    super(scope, id, props);
  }
}

```

JonHolman:
> > How about removing the without imports directory and parent.ts file? Since that's just calling our original, right?
> 
> The without imports directory just has a parent file. All it does it create an empty stack. Here's what it does:
> 
> ```ts
> import { Construct } from "constructs";
> import {
>   Stack,
>   StackProps,
> } from "aws-cdk-lib";
> import { DeploymentConfigProperties } from "../../deployment-config";
> 
> export class WithoutImportsParentStack extends Stack {
>   constructor(
>     scope: Construct,
>     id: string,
>     props: StackProps & DeploymentConfigProperties
>   ) {
>     super(scope, id, props);
>   }
> }
> ```
> 
> Going to go ahead and merge into 4184 branch now and if we need to adjust further we can rediscuss there as everyone is looking through what imports would mean and decide if we're going that route at all.

Ah.  I misunderstood.  Thank you.

PR #15026, opened 2025-01-09
CMDCT-4238 - Prevent destroy from destroying serverless stacks unless particular resources are set to retain

JonHolman:
### Description
Idea to avoid accidentally deleting important resources during the CDK migration

### Related ticket(s)
CMDCT-4238

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -2,6 +2,7 @@ import yargs from "yargs";
 import * as dotenv from "dotenv";
 import LabeledProcessRunner from "./runner.js";
 import { ServerlessStageDestroyer } from "@stratiformdigital/serverless-stage-destroyer";
+import { getCloudFormationTemplatesForStage } from "./getCloudFormationTemplatesForStage.js";
 import { execSync } from "child_process";
 
 // load .env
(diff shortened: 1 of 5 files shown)


src/run.ts
```diff
@@ -186,6 +190,82 @@ async function destroy_stage(options: {
...
+
+  if (
+    ["master", "val", "production"].includes(options.stage) &&
+    notRetained.length > 0
+  ) {
+    console.log(
+      "Will not destroy the stage because its an important stage and some important resources are not yet set to be retained:"
+    );
+    notRetained.forEach(({ templateKey, resourceKey }) =>
+      console.log(` - ${templateKey}/${resourceKey}`)
```
peoplespete, line 264: Sweet this is the beautiful thing!

src/run.ts
```diff
@@ -186,6 +190,82 @@ async function destroy_stage(options: {
...
+      resourceKey: "StateFormsTable",
+    },
+    { templateKey: `database-${options.stage}`, resourceKey: "StatesTable" },
+    { templateKey: `database-${options.stage}`, resourceKey: "AuthUserTable" },
+    {
+      templateKey: `ui-${options.stage}`,
+      resourceKey: "CloudFrontDistribution",
+    },
+    { templateKey: `ui-auth-${options.stage}`, resourceKey: "CognitoUserPool" },
+  ];
```
peoplespete, line 246: You'd have to change some other code I'm sure but I wonder if you like this stripped down version of this same data:
```js
const resourcesToCheck = {
  `database-${options.stage}`: "FormAnswersTable",
  `database-${options.stage}`: "FormQuestionsTable",
  `database-${options.stage}`: "FormTemplatesTable",
  `database-${options.stage}`: "FormsTable"
  `database-${options.stage}`: "StateFormsTable",
  `database-${options.stage}`,: "StatesTable"
  `database-${options.stage}`:  "AuthUserTable"
  `ui-${options.stage}`: "CloudFrontDistribution",
  `ui-auth-${options.stage}`: "CognitoUserPool" 
}
```

src/run.ts
```diff
@@ -186,6 +190,82 @@ async function destroy_stage(options: {
...
+    ({ templateKey, resourceKey }) => {
+      const policy =
+        templates?.[templateKey]?.Resources?.[resourceKey]?.DeletionPolicy;
+      return policy !== "Retain";
+    }
+  );
+
+  if (
+    ["master", "val", "production"].includes(options.stage) &&
+    notRetained.length > 0
```
peoplespete, line 258:
```suggestion
  if (
    ["master", "val", "production"].includes(options.stage) {
    # all the code that gets notRetained
      if (notRetained.length > 0) {
       # then the code to bail
```
This would make it so we only make the calls to check all this stuff if we are talking about an isDev stage.

src/run.ts
```diff
@@ -186,6 +190,82 @@ async function destroy_stage(options: {
...
+  const notRetained = resourcesToCheck.filter(
+    ({ templateKey, resourceKey }) => {
+      const policy =
+        templates?.[templateKey]?.Resources?.[resourceKey]?.DeletionPolicy;
+      return policy !== "Retain";
+    }
+  );
+
+  if (
+    ["master", "val", "production"].includes(options.stage) &&
```
peoplespete, line 257: Any way to make this match isDev's definition? Also if that's complicated feel free to ignore.

peoplespete (approved):
A few things to consider but when you're happy I'm happy

src/run.ts
JonHolman: This isn't in the cdk branches, this will be in sls branches, so we don't have isDev.

peoplespete (approved):
Chatted in slack, ready when you are

PR #15041, opened 2025-01-28
CMDCT-4247 - v4 serverless upgrade

peoplespete:
### Description
Upgrade to Serverless V4

### Related ticket(s)
CMDCT-4247

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui/serverless.yml
```diff
@@ -308,17 +314,6 @@ resources:
     DeliveryRole:
       Type: AWS::IAM::Role
```
JonHolman, line 315: This doesn't need the IAM Path and Permission Boundary?  I'm confused how the feature branch deployed.

services/ui-auth/libs/users.js
```diff
@@ -1,4 +1,4 @@
-[
```
JonHolman, line 1: We're keeping this as users.json in the other apps, right?

services/ui-auth/libs/cognito-lib.js
```diff
@@ -1,57 +1,23 @@
...
 
 export async function createUser(params) {
-  await new Promise((resolve, _reject) => {
-    COGNITO_CLIENT.adminCreateUser(params, function (err, _data) {
-      var response;
-      if (err) {
-        console.log("FAILED ", err, err.stack); // eslint-disable-line no-console
-        response = { statusCode: 500, body: { message: "FAILED", error: err } };
-        resolve(response); //if user already exists, we still continue and ignore
-      } else {
```
benmartin-coforma, line 17: In our other apps, when removing this code, we added try/catches around the call site in createUsers. That was done in the name of preventing this lambda from failing when users already exist - which I think might happen for the second, third, etc deployments in each ephemeral environment? Although I can't say I've ever seen it myself, and I'm not sure if that failure would have any adverse consequences (for all I know it would just send back a 500, which the deploy pipeline might be ignoring).

For an example of what I'm talking about: https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/ui-auth/handlers/createUsers.js#L25-L44

services/app-api/package.json
```diff
@@ -7,25 +7,24 @@
...
-    "serverless-offline-ssm": "^6.2.0"
+    "serverless-associate-waf": "^1.2.1"
   },
   "dependencies": {
     "@aws-sdk/client-dynamodb": "^3.709.0",
     "@aws-sdk/client-ses": "^3.709.0",
     "@aws-sdk/client-ssm": "^3.709.0",
     "@aws-sdk/lib-dynamodb": "^3.709.0",
     "aws-jwt-verify": "^4.0.1",
+    "jwt-decode": "^4.0.0",
```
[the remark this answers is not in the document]
peoplespete: Sweet. I looked and can't find it in the root package.json. In fact, before I added it I could only find it as a dependency of some other dependency in a yarn.lock so it might have just been finding that before.

run
```diff
@@ -32,9 +32,9 @@ if ! which yarn > /dev/null ; then
 fi
 
 # check serverless is installed globally.
-if ! which serverless > /dev/null ; then
-	echo "installing serverless globally"
-	yarn global add serverless@3.38.0
+if ! which serverless > /dev/null || [[ "$(serverless --version | cut -d'.' -f1)" != "4" ]]; then
+	echo "installing serverless v4 globally"
+	yarn global add serverless@4.4.18
```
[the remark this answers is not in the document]
peoplespete: Good catch! It varies yes but the thing was still broken :). Fixed.

services/ui-auth/libs/cognito-lib.js
peoplespete: Ah Good point! Fixed.

services/ui-auth/libs/users.js
peoplespete: Yep, my B!

peoplespete:
@benmartin-coforma @JonHolman Thanks for reviewing! I've made requested changes and it should be ready for a final looksee. 🙏

JonHolman (dismissed):

services/app-api/package.json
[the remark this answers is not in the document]
benmartin-coforma: Oh, you're right! Even better.

run
```diff
@@ -32,10 +32,10 @@ if ! which yarn > /dev/null ; then
...
 
 # check serverless is installed globally.
-if ! which serverless > /dev/null ; then
-	echo "installing serverless globally"
-	yarn global add serverless@3.38.0
-	export PATH="$PATH:$(yarn global bin)"
+if ! which serverless > /dev/null || ! serverless --version 2>&1 | grep -q " 4."; then
+    echo "installing serverless v4 globally"
+    yarn global add serverless@4.4.18
+    export PATH="$PATH:$(yarn global bin)"
```
benmartin-coforma, line 38: I think this will fire false positives. Note that my output from v3 had three lines:
```
Framework Core: 3.33.0
Plugin: 6.2.3
SDK: 4.3.2

```
And that *does* contain the substring `" 4."`!

It's super annoying that they changed the format of their `--version` output. One of the primary purposes of that command is our use case right here - trying to figure out programmatically if an upgrade is needed!

Let's make the grep more specific. Maybe this?
```suggestion
if ! which serverless > /dev/null || ! serverless --version 2>&1 | grep -q "Framework 4."; then
    echo "installing serverless v4 globally"
    yarn global add serverless@4.4.18
    export PATH="$PATH:$(yarn global bin)"
```

I tested it locally, where I have a global v3 install, and a temp project with a v4 install, and it seemed to do the trick:
image

benmartin-coforma (changes requested):

benmartin-coforma (approved):

PR #15055, opened 2025-02-11
CMDCT-4328 - standardizes the way we use env variables to access dynamo tables

peoplespete:
### Description
This PR standardizes the way we use env variables to access dynamo tables

### Related ticket(s)
CMDCT-4328

---
### How to test
Run app! It should still work :)

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/stacks/api.ts
+++ b/deployment/stacks/api.ts
@@ -114,10 +114,7 @@ export function createApiComponents(props: CreateApiComponentsProps) {
     BOOTSTRAP_BROKER_STRING_TLS: brokerString,
     stage,
     ...Object.fromEntries(
-      tables.map((table) => {
-        const tableNameEnvCase = table.id.split(/\.?(?=[A-Z])/).join('_').toUpperCase() // FormAnswers -> FORM_ANSWERS
-        return [`${tableNameEnvCase}_TABLE`, table.name]
-      })
+      tables.map((table) => [`${table.id}Table`, table.name])
     ),
   };
 
--- a/deployment/stacks/api.ts
+++ b/deployment/stacks/api.ts
@@ -114,7 +114,10 @@ export function createApiComponents(props: CreateApiComponentsProps) {
     BOOTSTRAP_BROKER_STRING_TLS: brokerString,
     stage,
     ...Object.fromEntries(
-      tables.map((table) => [`${table.id}TableName`, table.name])
+      tables.map((table) => {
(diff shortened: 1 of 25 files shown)


deployment/stacks/api.ts
```diff
@@ -114,7 +114,10 @@ export function createApiComponents(props: CreateApiComponentsProps) {
     BOOTSTRAP_BROKER_STRING_TLS: brokerString,
     stage,
     ...Object.fromEntries(
-      tables.map((table) => [`${table.id}TableName`, table.name])
+      tables.map((table) => {
+        const tableNameEnvCase = table.id.split(/\.?(?=[A-Z])/).join('_').toUpperCase() // FormAnswers -> FORM_ANSWERS
+        return [`${tableNameEnvCase}_TABLE`, table.name]
+      })
```
benmartin-coforma, line 120: The regex here is a little magical, which is unfortunate. The comment is helpful, but I'm still tempted to add a property to our `DynamoDBTableIdentifiers` object for `envVarName`, and list them all explicitly... it would be ten extra lines of config code, to remove this bit of magic here. But it would also make our setup easier for newbies to grep, which is worth something too.

(I can tell the regex is magic because the `\.?` trigraph is doing nothing at all - since none of our table names contain dots 😁)

benmartin-coforma (approved):
This is an improvement for sure! I did make a temp branch that contains both this code and the tests from #15048 , and confirmed that all the tests passed*

My comment below is a suggestion, but not necessarily blocking. What do you think?

*Once I made appropriate updates to setupTests.js, of course, and updated the expected number of scan calls from forceKafkaSync.js

deployment/stacks/api.ts
benmartin-coforma: 🤔  I suppose alternatively we could just use the PascalCase names as is, and instead remove the SCREAMING_SNAKE_CASE references from the api code. It's a little unfortunate that that bucks the usual environment variable naming trend, but if it makes our setup code simpler, that might be a good tradeoff.

peoplespete: I thought of one more option which is to make the table name initially be in SCREAMING_SNAKE_CASE but then it doesn't match the convention of all the other resources in our cloudformation stack.
So...
I'm going to do PascalCase as you suggested. Seems worth breaking convention to have everything simpler and remove the dirty regex magic. 👍

peoplespete: Done.

PR #15058, opened 2025-02-12
CMDCT-4291 - localstack for local development

peoplespete:
### Description
Adds support for localstack

### Related ticket(s)
CMDCT-4291

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


deployment/constructs/lambda.ts
```diff
@@ -102,7 +102,10 @@ export class Lambda extends Construct {
         method,
         new apigateway.LambdaIntegration(this.lambda),
         {
-          authorizationType: apigateway.AuthorizationType.IAM,
+          authorizationType:
+            process.env.CDK_DEFAULT_ACCOUNT !== "000000000000"
+              ? apigateway.AuthorizationType.IAM
+              : undefined,
```
peoplespete, line 108:
```suggestion
          ...(process.env.CDK_DEFAULT_ACCOUNT !== "000000000000" && { authorizationType: apigateway.AuthorizationType.IAM })
```

I recently discovered this sort of thing is possible to conditionally add keys. Do you like it or think it's terrible? Part of me likes it but part of me thinks it isn't that readable.

deployment/constructs/waf.ts
```diff
@@ -16,142 +13,141 @@ interface WafProps {
...
-    scopeType: string,
+    scopeType: string
   ) {
     super(scope, id);
 
-    const {
-      name,
-      blockRequestBodyOver8KB = true
-    } = props;
+    if (process.env.CDK_DEFAULT_ACCOUNT !== "000000000000") {
```
peoplespete, line 27: I wonder if we could hide this logic and just import a constant that is named something like:
`isLocalStack` or something

deployment/stacks/api.ts
```diff
@@ -442,10 +442,12 @@ export function createApiComponents(props: CreateApiComponentsProps) {
     },
     "REGIONAL"
   );
-  new wafv2.CfnWebACLAssociation(scope, "WebACLAssociation", {
-    resourceArn: api.deploymentStage.stageArn,
-    webAclArn: waf.webAcl.attrArn,
-  });
+  if (waf.webAcl) {
```
peoplespete, line 445: Is that true only for non-localstack? If so I wonder if it'd be more clear to do the `isLocalStack` check vs this one. Or even just a little comment explaining would suffice.

deployment/stacks/api.ts
```diff
@@ -455,12 +457,14 @@ export function createApiComponents(props: CreateApiComponentsProps) {
       enforceSSL: true,
     });
 
-    new CloudWatchToS3(scope, "CloudWatchToS3Construct", {
-      logGroup: waf.logGroup,
-      bucket: logBucket,
-      iamPermissionsBoundary: props.iamPermissionsBoundary,
-      iamPath: props.iamPath,
-    });
+    if (waf.logGroup) {
```
peoplespete, line 460: Same as above.

deployment/stacks/ui-auth.ts
```diff
@@ -262,10 +262,12 @@ export function createUiAuthComponents(props: CreateUiAuthComponentsProps) {
     "REGIONAL"
   ).webAcl;
 
-  new wafv2.CfnWebACLAssociation(scope, "CognitoUserPoolWAFAssociation", {
-    resourceArn: userPool.userPoolArn,
-    webAclArn: webAcl.attrArn,
-  });
+  if (webAcl) {
```
peoplespete, line 265:
```suggestion
  if (!isLocalStack) {
```

services/app-api/libs/authorization.js
```diff
@@ -7,29 +7,41 @@ import { SimpleJsonFetcher } from "aws-jwt-verify/https";
 
 export async function getUserDetailsFromEvent(event) {
```
peoplespete, line 8: How would you feel about this change to getUserDetailsFromEvent?
```js
export async function getUserDetailsFromEvent(event) {
  await verifyEventSignature(event);

  if (event.requestContext.accountId === "000000000000") { 
    return {}
  }

  // TODO, it seems that jwtDecode and verifier.verify may return the same object?
  // Maybe we can remove the jwtDecode dependency.

  const token = jwtDecode(event?.headers?.["x-api-key"]);
  const role = mapMembershipToRole(token["custom:ismemberof"]);

  return {
    email: token.email,
    firstName: token.given_name,
    lastName: token.family_name,
    role,
    username: token.identities?.[0]?.userId || token.email,
    usernameSub: token.sub, // whatever that means.
  };
}
```

src/run.ts
```diff
@@ -206,11 +315,15 @@ async function destroy({
 // All valid arguments to dev should be enumerated here, this is the entrypoint to the script
 yargs(process.argv.slice(2))
   .command(
-    "local",
+    "watch",
```
peoplespete, line 318: Oh neato, is this so that we can use cdk-watch and local separately to compare?

JonHolman: originally I deleted the cdk watch stuff.  This is to not delete the cdk watch stuff.

src/run.ts
```diff
@@ -103,11 +105,102 @@ async function run_cdk_watch(runner: LabeledProcessRunner, options: { stage: str
...
+  try {
+    const output = execSync("localstack status", {
+      encoding: "utf-8",
+      stdio: "pipe",
+    }).trim();
+    return output.includes("running");
+  } catch {
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
JonHolman: I don't have pipx on my machine.  
maybe it makes things more efficient, but you don't need to pass that service list before localstack start

Here's my first attempt:

https://app.localstack.cloud/sign-up

brew install localstack/tap/localstack-cli
npm install -g aws-cdk-local aws-cdk
pip install awscli-local

install docker

localstack start

./run local

Useful URL to monitor local localstack instance: https://app.localstack.cloud/inst/default/status

services/app-api/libs/authorization.js
JonHolman: everything I did so far to x-api-key was just awful, I need to work on it more and figure out why I had to do that to get login to work, what I did was wrong, very wrong.

deployment/stacks/ui-auth.ts
JonHolman: Well...  I don't like what I did, but I don't like that either, I don't want to do either.  So what happened is, we don't have WAF with localstack because we don't have pro.  We can either if every place we call the WAF construct or I thought it would be more efficient to just make the WAF construct do nothing when its localstack...  well...  then the WAF construct's properties need to have an undefined possibility...  that's why this if (webAcl) needs to be here to make sure its not undefined.  If we don't have to add the undefined potential, we don't need to add this if.  It may be better to huddle to discuss.

deployment/stacks/api.ts
JonHolman: same as below

deployment/stacks/api.ts
JonHolman: Yes, with localstack because we don't have pro. We can either if every place we call the WAF construct or I thought it would be more efficient to just make the WAF construct do nothing when its localstack... well... then the WAF construct's properties need to have an undefined possibility... that's why this if (waf.webAcl) needs to be here to make sure its not undefined. If we don't have to add the undefined potential, we don't need to add this if. It may be better to huddle to discuss.

deployment/constructs/waf.ts
JonHolman: sounds good!

deployment/constructs/lambda.ts
JonHolman: Yeah, I think that's less readable.  I'm flexible though.

peoplespete: Cool cool 👍

deployment/stacks/parent.ts
```diff
@@ -68,63 +68,72 @@ export class ParentStack extends Stack {
...
-      userPoolClientId,
-    } = createUiAuthComponents({
-      ...commonProps,
-      oktaMetadataUrl,
-      applicationEndpointUrl,
-      restApiId,
-      bootstrapUsersPasswordArn,
-      customResourceRole,
-    });
+    if (process.env.CDK_DEFAULT_ACCOUNT !== "000000000000") {
```
peoplespete, line 71: How would you feel about putting this logic into it's own file that looks like this:
```js
export function isLocalstack() {
  return process.env.CDK_DEFAULT_ACCOUNT == "000000000000"
}
```
That way we could import that and make it a tad more readable so it'd be like:

```js
if (!isLocalstack()) {
  //do something
}
```

src/run.ts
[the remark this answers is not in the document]
peoplespete: Cool cool. Yeah this was very much meant to be a take the good and chuck the bad situation. 👍

src/run.ts
```diff
@@ -103,11 +106,126 @@ async function run_cdk_watch(runner: LabeledProcessRunner, options: { stage: str
...
+    (output) => output.OutputKey === "SeedDataFunctionName"
+  )?.OutputValue;
+
+  const lambdaClient = new LambdaClient({ region: "us-east-1" });
+  const lambdaCommand = new InvokeCommand({
+    FunctionName: seedDataFunctionName,
+    InvocationType: "Event",
+    Payload: Buffer.from(JSON.stringify({})),
+  });
+  await lambdaClient.send(lambdaCommand);
```
peoplespete, line 217: I wonder if we could put this in our cdk code like we did for bootstrapUsers. We could add the only-for-localstack-if around it if that is needed.

JonHolman: :) this is in our CDK code, I had to convert it back right here only for localstack because we don't have localstack pro :)

deployment/stacks/parent.ts
JonHolman: yeah, you said this. https://github.com/Enterprise-CMCS/macpro-mdct-seds/pull/15058#discussion_r1953246987

sounds good.

src/write-ui-env-file.ts
```diff
@@ -12,7 +12,7 @@ const configFilePath = path.resolve(outputPath, "env-config.js");
 const project = "seds";
 const region = "us-east-1";
 
-export async function writeUiEnvFile(stage: string, local = false) {
+export async function writeLocalUiEnvFile(stage: string) {
```
peoplespete, line 15: As I'm thinking this through. I think we could add some conditional logic related to `isLocalstack` and then not need this file at all. The cdk creation of the ssm parameter will be correct regardless of it's context.

peoplespete: In fact I think you already sort of did it or most of it. I'll comment there.

JonHolman: this code is to create the env-config.js file for the react frontend.

deployment/stacks/parent.ts
```diff
@@ -68,63 +68,72 @@ export class ParentStack extends Stack {
...
+          applicationEndpointUrl,
+          s3BucketName,
+          cloudfrontDistributionId,
+          identityPoolId,
+          userPoolId,
+          userPoolClientId,
+          userPoolClientDomain: `${userPoolDomainName}.auth.${this.region}.amazoncognito.com`,
+        }),
+        description: `Deployment output for the ${stage} environment.`,
+      });
```
peoplespete, line 123: Since this is inside a localstack context, do we need to adjust these values as you did here: https://github.com/Enterprise-CMCS/macpro-mdct-seds/pull/15058/files#diff-06fcc7a4ca1aadf07dacee3ad767f6ee8ce9d6acc517266db5932125636a45d1R33

deployment/constructs/lambda.ts
```diff
@@ -102,7 +103,9 @@ export class Lambda extends Construct {
         method,
         new apigateway.LambdaIntegration(this.lambda),
         {
-          authorizationType: apigateway.AuthorizationType.IAM,
+          authorizationType: !isLocalStack()
+            ? apigateway.AuthorizationType.IAM
+            : undefined,
```
peoplespete, line 108:
```suggestion
          authorizationType: isLocalStack()
            ? undefined
            : apigateway.AuthorizationType.IAM,
```
Just reversing to remove the not
Optional nitpick

deployment/local/prerequisites.ts
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    Tags.of(mdctSubnet1).add("Name", "mdct-subnet-1");
+  }
+}
+
+async function main() {
+  const app = new App();
+
+  new LocalPrerequisiteStack(app, "seds-LocalPrerequisites");
```
peoplespete, line 32:
```suggestion
  new LocalPrerequisiteStack(app, "seds-local-prerequisites");
```
Another nitpick to match how seds-prerequisites is cased

deployment/local/util.ts
```diff
@@ -0,0 +1,3 @@
+export function isLocalStack() {
+  return process.env.CDK_DEFAULT_ACCOUNT === "000000000000";
+}
```
peoplespete, line 3: There's a chance this could be:
export const isLocalStack = process.env.CDK_DEFAULT_ACCOUNT === "000000000000";
but again I'm having to look really hard to find any improvements

deployment/stacks/api.ts
```diff
@@ -433,34 +434,39 @@ export function createApiComponents(props: CreateApiComponentsProps) {
...
+      });
+
+      if (waf.logGroup) {
+        new CloudWatchToS3(scope, "CloudWatchToS3Construct", {
+          logGroup: waf.logGroup,
+          bucket: logBucket,
+          iamPermissionsBoundary: props.iamPermissionsBoundary,
+          iamPath: props.iamPath,
+        });
+      }
```
peoplespete, line 468:
```suggestion
      new CloudWatchToS3(scope, "CloudWatchToS3Construct", {
        logGroup: waf.logGroup,
        bucket: logBucket,
        iamPermissionsBoundary: props.iamPermissionsBoundary,
        iamPath: props.iamPath,
      });
```
Shouldn't need this if anymore iirc

services/app-api/libs/authorization.js
```diff
@@ -6,7 +6,9 @@ import { SimpleJwksCache } from "aws-jwt-verify/jwk";
 import { SimpleJsonFetcher } from "aws-jwt-verify/https";
 
 export async function getUserDetailsFromEvent(event) {
-  await verifyEventSignature(event);
+  if (event.requestContext.accountId !== "000000000000") {
```
peoplespete, line 9: Can this use `isLocalstack` or is it not a thing to share lambda and cdk code?
If it isn't we could recreate the helper or even just a comment so future devs get that it's a localstack thing.

src/run.ts
```diff
@@ -84,14 +85,23 @@ function updateEnvFiles() {
...
 // run_fe_locally runs the frontend and its dependencies locally
 // @ts-ignore
-async function run_fe_locally(runner: LabeledProcessRunner, options: { stage: string }) {
-  const stage = options.stage;
-  await writeUiEnvFile(stage, true);
+async function run_fe_locally(runner: LabeledProcessRunner) {
+  const apiUrl = await getCloudFormationStackOutputValue(
+    "seds-jon-cdk",
+    "ApiUrl"
+  );
```
peoplespete, line 92: Wouldn't this mean that the local app is calling to the api of `seds-jon-cdk`?
I could be confused!

src/run.ts
```diff
@@ -103,11 +113,139 @@ async function run_cdk_watch(runner: LabeledProcessRunner, options: { stage: str
...
+    "--no-rollback",
+  ];
+  await runner.run_command_and_output("CDK deploy", deployCmd, ".");
+
+  const seedDataFunctionName = await getCloudFormationStackOutputValue(
+    "seds-jon-cdk",
+    "SeedDataFunctionName"
+  );
+
+  if (seedDataFunctionName) {
```
peoplespete, line 229: Maybe I'm too bold here but I think we can assume this is there.

src/run.ts
```diff
@@ -84,14 +85,23 @@ function updateEnvFiles() {
...
-async function run_fe_locally(runner: LabeledProcessRunner, options: { stage: string }) {
-  const stage = options.stage;
-  await writeUiEnvFile(stage, true);
+async function run_fe_locally(runner: LabeledProcessRunner) {
+  const apiUrl = await getCloudFormationStackOutputValue(
+    "seds-jon-cdk",
+    "ApiUrl"
+  );
+
+  if (apiUrl) {
```
peoplespete, line 94: Maybe I'm too bold here but I think we can assume this is there.

JonHolman: I agree, but TypeScript disagrees.  Should we use `!` below to get around it?

src/run.ts
JonHolman: I agree, but TypeScript disagrees.  Should we use `!` below to get around it?

src/run.ts
peoplespete: I like that!

src/run.ts
peoplespete: Ah ok. Can we change the automatically named local stack's name to be `local` or something?
Something like this:
```ts
  const deployCmd = [
    "cdklocal",
    "deploy",
    "--context",
    "stage=local",
    "--all",
    "--no-rollback",
  ];
```

peoplespete: Then that would make more sense to me when we look for the output from the stack

services/app-api/libs/authorization.js
JonHolman: It would be complicated to share code between our lambdas and CDK.  For now I just recreated the isLocalStack within this function.  If we need it in another lambda, we can export it in a utility.

JonHolman (approved):

PR #15062, opened 2025-02-19
CMDCT-4337 - adds colima for docker for localstack + updating documentation

peoplespete:
### Description
Updates docs and introduces colima

### Related ticket(s)
CMDCT-4337

---
### How to test
Try to follow instructions here: https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/cmdct-4337/deployment/local/README.md

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/local/README.md
+++ b/deployment/local/README.md
@@ -10,29 +10,12 @@ Before running the application locally, ensure the following dependencies are in
 
 1. **Colima/Docker** - LocalStack runs inside a Colima container that uses docker as it's runtime.
 
-Links:
+_The install is handled by the run script._
+
+Links for the curious:
 
 - Docker - https://www.docker.com/get-started
 - Colima - https://github.com/abiosoft/colima
-
-```sh
-# Install Docker
-brew install docker
-# now this should print something out:
-docker -v
-
-# Install Colima
-brew install colima
-# now this should print something out:
-colima --version
-# this should keep colima running better when you log on and off:
-brew services start colima
-# start colima with specific vm-type (this prevents crashing which shows up as "socket hang up")
-colima start --vm-type=vz
-# verify colima is started
-colima status
-# should now include the phrase "colima is running"
-```
 
 Now add this line to the bottom of your bash/zsh rc/profile:
 This is probably the file: `~/.zprofile` if you're using the standard mac setup.
@@ -46,23 +29,11 @@ Close and reopen your terminal.
 
 2. **LocalStack** - Provides a local AWS emulating environment.
 
-   - Sign up for a free account: [LocalStack Cloud](https://app.localstack.cloud/sign-up)
-   - Install LocalStack CLI:
-     ```sh
-     brew install localstack/tap/localstack-cli
-     ```
+_The install is handled by the run script._
 
 3. **AWS CLI Local** - Required for interacting with LocalStack.
 
-   ```sh
-   brew install pipx
-   pipx install awscli-local
-   ```
-
-4. **AWS CDK Local**
-   ```sh
-   npm install -g aws-cdk-local aws-cdk
-   ```
+_The install is handled by the run script._
 
 ## Running LocalStack
 
@@ -70,6 +41,7 @@ Start the LocalStack service before deploying the application:
 
 ```sh
 localstack start
+# this will keep running during local development
 ```
 
 ## Deploying Locally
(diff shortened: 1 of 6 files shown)


run
```diff
@@ -31,13 +31,6 @@ if ! which yarn > /dev/null ; then
 	exit 1
 fi
 
-# check serverless is installed globally.
```
JonHolman, line 34: Nice!

run
```diff
@@ -47,32 +40,54 @@ fi
...
         ;;
     esac
+    ARGS+=("$arg")
 done
 
-# Inject secrets to .env if the flag is set
 if [ "$UPDATE_ENV" = true ]; then
-    # Check if 1Password CLI is installed
     if ! which op > /dev/null; then
-        echo "The MDCT team uses 1Password to store and retrieve development credentials. If you do not have a 1Password account please request one."
```
JonHolman, line 67: We should still have this note to request a 1password account if you don't have one, right?

run
```diff
@@ -47,32 +40,54 @@ fi
...
         exit 1
     fi
-
-    # call 1password vault to retrieve secrets
-    op inject -i .env.tpl -o .env -f
-    sed -i "" -e "s/# pragma: allowlist secret//g" .env
 fi
 
+if [ "$LOCAL" = true ]; then
+    if ! which docker > /dev/null ; then
```
JonHolman, line 66: We don't need colima and docker installed.

run
```diff
@@ -47,32 +40,54 @@ fi
 
 # Parse arguments
 UPDATE_ENV=false
+LOCAL=false
 ARGS=()
 
 for arg in "$@"; do
     case $arg in
-        --update-env)
+        update-env)
```
JonHolman, line 48: I believe we could also run `./run local --update-env` that will still work, right?

deployment/local/README.md
```diff
@@ -8,53 +8,71 @@ Before running the application locally, ensure the following dependencies are in
...
-   ```sh
-   npm install -g aws-cdk-local aws-cdk
-   ```
+Now add this line to the bottom of your bash/zsh rc/profile:
+This is probably the file: `~/.zprofile` if you're using the standard mac setup.
 
-## Open Docker
+```sh
+# this tells docker that you're using colima and to look at colima to answer questions like: are any containers running?
+export DOCKER_HOST="unix://${HOME}/.colima/default/docker.sock"
```
JonHolman, line 25: I don't have this and I've been fine.

JonHolman (approved):
Some feedback, but feel free to merge when you're good with it.

deployment/local/README.md
peoplespete: So one thing to check is if you are really using colima. Basically start up localstack and then look at docker (`docker ps`). If you see a container, you're not actually using colima.

run
peoplespete: So I removed this because it was doing the same thing as `./run update-env` and it didn't feel like we needed two ways to update the env. If this is a dealbreaker let me know and I can try and find a way to add it back. I am just sort of on a quest so that if you want to update-env there is one and only one way to do that.

run
peoplespete: My 2c is we should either be specific or remove it. Like what is an email address they should send a message to to request a 1password account? As is we're being too vague to be helpful to a real person. What do you think?

run
peoplespete: I tried to run colima without docker and it didn't work. Let me know if you manage that and I'll change all this.

peoplespete: This is what I get:
```
❌ Error: Docker could not be found on the system.
Please make sure that you have a working docker environment on your machine.
```

deployment/local/README.md
JonHolman: I don't know if that's true.  Look at the demo at the top of their readme at https://github.com/abiosoft/colima

They wouldn't use docker ps to demo if it wasn't their tool behind the scenes.

run
JonHolman: I don't think its a big deal either way.  I think the guidance would have to be if you don't have a 1password account, ask your on-boarding lead, which is what I hope they will do if we leave them hanging with no guidance :)

run
JonHolman: Not a problem for me, but some on the team may say something because this may be functionality they're used to using.

peoplespete: Sounds good. I'll manage that fall out :)

run
peoplespete: Yeah my thoughts exactly. I'll leave it out for now and if people get hung up we can hassle some folks for a more specific message.

deployment/local/README.md
peoplespete: Ok this is great news. So we'll just use the brew install docker version and not have license issues and I'll remove this whole bit.

run
peoplespete: I didn't understand this. Now I do. It's cool now. We will have docker cli and colima installed together. As long as we avoid docker desktop we don't need licenses.

PR #15063, opened 2025-02-20
CMDCT-4338 one secret to rule them all

peoplespete:
### Description
Streamlines the use of secrets by the cdk version of SEDS

### Related ticket(s)
CMDCT-4338

---
### How to test
Just make sure the app didn't break :)
Mainly bootstrapUsers should have worked which means you can still log in

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---


deployment/import_instructions.md
peoplespete, line 1: Ignore this. I change the capitalization and it took on jon-cdk (or is about to have) but it didn't remove the old file so I'm doing it in this PR.

peoplespete: git is really weird about only changing the capitalization of a file name actually

services/ui-auth/handlers/createUsers.js
```diff
@@ -23,7 +15,7 @@ export async function handler(_event, _context, _callback) {
       UserAttributes: user.attributes,
     };
     var passwordData = {
-      Password: password,
+      Password: process.env.bootstrapUsersPassword,
```
peoplespete, line 18: This replicates how it works on master right now.

deployment/import_instructions.md
benmartin-coforma: It 1000% is. At a previous job I was haunted for years by a miscapitalized directory name I just couldn't seem to permanently delete, no matter what I did.

deployment/prerequisites.ts
```diff
@@ -39,9 +37,9 @@ export class PrerequisiteStack extends Stack {
         permissionsBoundary: iam.ManagedPolicy.fromManagedPolicyArn(
           this,
           "iamPermissionsBoundary",
-          iamPermissionsBoundaryArn
+          "arn:aws:iam::${AWS::AccountId}:policy/cms-cloud-admin/developer-boundary-policy"
```
JonHolman, line 40: Right now this doesn't exist in our local stack setup, we could add it to our local stack pre reqs.  Thoughts?

deployment/stacks/parent.ts
```diff
@@ -24,24 +24,12 @@ export class ParentStack extends Stack {
...
-      isDev,
-      vpcName,
-      bootstrapUsersPasswordArn,
-      oktaMetadataUrl,
-      brokerString,
-      iamPermissionsBoundaryArn,
-      iamPath,
-      deploymentConfigParameters,
-    } = props;
+    const iamPermissionsBoundaryArn = `arn:aws:iam::${process.env.CDK_DEFAULT_ACCOUNT}:policy/cms-cloud-admin/developer-boundary-policy`
```
JonHolman, line 27: In other places we're using `${Aws.ACCOUNT_ID}` when I looked into this before, I landed on that as a preference.  If you disagree, we can switch, but I would like us to go all in on one or the other.

deployment/stacks/parent.ts
```diff
@@ -50,7 +38,7 @@ export class ParentStack extends Stack {
       iamPath,
     };
 
-    const vpc = ec2.Vpc.fromLookup(this, "Vpc", { vpcName });
+    const vpc = ec2.Vpc.fromLookup(this, "Vpc", { vpcName: commonProps.vpcName });
```
JonHolman, line 41: this is a nit, but why commonsProps and not props and why did we remove the deconstruction assigning const vpcName

deployment/stacks/ui-auth.ts
```diff
@@ -18,26 +18,27 @@ interface CreateUiAuthComponentsProps {
...
   isDev: boolean;
-  oktaMetadataUrl: string;
   applicationEndpointUrl: string;
   restApiId: string;
-  bootstrapUsersPasswordArn: string;
-  iamPermissionsBoundary: IManagedPolicy;
-  iamPath: string;
   customResourceRole: iam.Role;
+  iamPath: string;
+  iamPermissionsBoundary: IManagedPolicy;
```
JonHolman, line 25: We're hardcoding this and iamPath now, right?  If so, why do we still have as props?

deployment/stacks/parent.ts
peoplespete: Fixed 👍

deployment/prerequisites.ts
peoplespete: Looks like you fixed this in your PR. Thanks!

peoplespete:
> We don't need to check that colima and docker are running, we just need one or the other.

Looks like you fixed this in your PR as well. Thanks!

deployment/stacks/ui-auth.ts
JonHolman: oh, gotcha.  either way.  Maybe instead of passing it in, we could just const iamPermissionsBoundary at the beginning and reuse many times to only have the ugly string once.

PR #15065, opened 2025-02-24
Fixing the localstack error "Cannot read properties of undefined (reading 'split')"

JonHolman:
### Description
Fixing the localstack error "Cannot read properties of undefined (reading 'split')"

### Related ticket(s)

CMDCT-4340

---
### Notes

---
### Pre-review checklist

- [X] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [X] I have updated relevant documentation, if necessary
- [X] I have performed a self-review of my code
- [X] I have manually tested this PR in the deployed cloud environment

---
#### Review

#### Security

---


benmartin-coforma (changes requested):
Sorry, I have a different thought here - give me a little time to write it up.

services/app-api/libs/kafka-source-lib.js
```diff
@@ -105,6 +105,10 @@ class KafkaSourceLib {
   }
 
   async handler(event) {
+    if (process.env.BOOTSTRAP_BROKER_STRING_TLS === "localstack") {
+      return;
+    }
```
peoplespete, line 110: I'm a little tempted to do something like what we did in `getUserDetailsFromEvent` like this:
```
  const isLocalStack = event.requestContext.accountId === "000000000000";
  if (!isLocalStack) {
    return;
  }
``` 
just so we are consistent. We could also create the kafka stuff via our local prerequisites stack but 🤷 that might get complicated.

peoplespete: But this is also probably fine since you don't have to pass the account id to kafka function

peoplespete (approved):
Cool cool

services/app-api/libs/kafka-source-lib.js
JonHolman: checking the account id for consistency makes sense, but lets see what @benmartin-coforma is thinking.

benmartin-coforma: Thanks for waiting on me! I spent some time trying to flesh out my idea, but I couldn't get it to a nice place.

In essence, I wanted to take the work that's done eagerly at the top of the file (creating the kafka instance, constructing the producer, and setting up the signal traps), and combine it with the work being done lazily down here (telling the producer to connect), into one lazy function. I also wanted to move this "are we local? if so, no-op" check into the individual callers (postKafkaData and dataConnectSource).

I still think that approach _could_ get us to a nicer spot overall? Especially because I don't think the environment variable access here is awesome; ideally the brokers would be passed in via constructor from the app code, rather than pulled out of the environment by the lib file. But ultimately, that would mean a more serious overhaul of the code, which isn't a good idea right now.

Maybe I'll still try a refactor like that down the road... but in the meantime, this change doesn't put us in a worse place, so I'm cool with it.

As for whether to read from the broker string or the account ID? Whichever you prefer is fine with me. I'm going to approve as-is, but consider any changes of that nature to have my pre-approval (as I'm on PTO for the next 2 days)

benmartin-coforma (approved):

benmartin-coforma:
Oh, and CodeClimate can stuff it. We're much better off with this code in one 360-line file than split across multiple files.

services/app-api/libs/kafka-source-lib.js
JonHolman: @peoplespete I looked at doing something like this instead:
```
    const isLocalStack = event.requestContext.accountId === "000000000000";
    if (isLocalStack) {
      return;
    }
```

but we will still need to do something with the environment variable BOOTSTRAP_BROKER_STRING_TLS because of line 7 of this file which is outside of the handler.  Since its outside of the handler, the event object is not available to it, so it doesn't know the account number.  Or we need to change line 7.  @peoplespete  thoughts?

PR #15083, opened 2025-03-10
CMDCT-4390 - fixes kafka getting non-existent tables

peoplespete:
### Description
This will fix a kafka problem. Hopefully that's all we need to do.

### Related ticket(s)
CMDCT-4390

---
### How to test
Check postKafta and dataConnect logs to see there is no error.

### Notes
Also, super uncool but I had to change the version for cdk because the internet is a spooky place.
How does this work?:
https://github.com/aws/aws-cdk/releases
Says the latest version is 2.182.0
image

Then check out yarn's versions
https://yarnpkg.com/package?q=aws-cdk&name=aws-cdk
image
There they've removed 2.182.0 and added several with the middle number in the thousands. The thousands introduces some things we'll need to change.

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/deployment-config.ts
+++ b/deployment/deployment-config.ts
@@ -15,6 +15,7 @@ export interface DeploymentConfigProperties {
   vpnIpSetArn?: string;
   vpnIpv6SetArn?: string;
   brokerString: string;
+  kafkaAuthorizedSubnetIds: string;
 }
 
 export const determineDeploymentConfig = async (stage: string) => {
@@ -73,6 +74,7 @@ function validateConfig(config: {
     "vpcName",
     "oktaMetadataUrl",
     "brokerString",
+    "kafkaAuthorizedSubnetIds",
   ];
 
   const invalidKeys = expectedKeys.filter(
--- a/deployment/stacks/api.ts
+++ b/deployment/stacks/api.ts
@@ -15,6 +15,7 @@ import {
 import { Lambda } from "../constructs/lambda";
 import { WafConstruct } from "../constructs/waf";
 import { addIamPropertiesToBucketAutoDeleteRole } from "../utils/s3";
+import { getSubnets } from "../utils/vpc";
 import { LambdaDynamoEventSource } from "../constructs/lambda-dynamo-event";
 import { DynamoDBTableIdentifiers } from "../constructs/dynamodb-table";
 import { isDefined } from "../utils/misc";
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
   "devDependencies": {
     "@types/node": "^20.12.7",
     "@types/yargs": "^15.0.10",
-    "aws-cdk": "^2.182.0",
+    "aws-cdk": "2.179.0",
```
JonHolman, line 33: huh?

benmartin-coforma (dismissed):
Whatever you want to do about the version # for now is fine with me.

package.json
peoplespete: See the notes on the PR. It's the thing we ran into when we were deploying on Sunday. yarn doesn't have a record of the version we used to use. 🤷

JonHolman (dismissed):

deployment/deployment-config.ts
```diff
@@ -15,6 +15,7 @@ export interface DeploymentConfigProperties {
   vpnIpSetArn?: string;
   vpnIpv6SetArn?: string;
   brokerString: string;
+  kafkaSubnetIds: string;
```
JonHolman, line 18: Should this be kafka authororized subnet ids?  This isn't actually Kafka, but the subnets allowed to get to kafka.

deployment/stacks/api.ts
```diff
@@ -15,6 +15,7 @@ import {
 import { Lambda } from "../constructs/lambda";
 import { WafConstruct } from "../constructs/waf";
 import { addIamPropertiesToBucketAutoDeleteRole } from "../utils/s3";
+import { getSubnets } from "../utils/vpc";
```
JonHolman, line 18: Shouldn't the VPC info and the subnet info come into the stack and "travel" together?  Meaning VPC name and subnet IDs.

deployment/deployment-config.ts
peoplespete: Changed to kafkaAuthorizedSubnetIds

deployment/stacks/api.ts
peoplespete: I could make it a more complicated nested thing but the vpc name isn't necessary to get the subnets if I have the ids. I really don't care unless we're going to do some nifty checks to see if the subnets are in the vpc but since we are configuring I think we should trust ourselves. If we went a name based filtering of subnets inside the vpc I'd be more inclined to.

BearHanded (approved):

benmartin-coforma (approved):

deployment/stacks/api.ts
peoplespete: We huddled and changes were made. 🙏

JonHolman (approved):

JonHolman (approved):

PR #15086, opened 2025-03-12
CMDCT-4403 - fix waf rule exclusion 8kb failure to allow

peoplespete:
### Description
Waf rule should be allowing 8kb payloads for app-api waf. It isn't. Now it should.

### Related ticket(s)
CMDCT-4403

---
### How to test
Make a really large form payload and save form and it will work now.

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---


benmartin-coforma (commented):
Isn't this backwards? I thought "count" means "this rule passes, but a subsequent rule might block this request", whereas "allow" means "let this request through immediately, no further rules are required". So wouldn't putting "allow" here potentially disable other protections?

I think the real fix is just that the boolean is inverted. The AWS Common rule set blocks large request bodies, so if we don't want them blocked, that's when we need the "count" override.

benmartin-coforma (approved):
According to my imperfect understanding of AWS WAFS, this change will do the trick 👍

PR #15087, opened 2025-03-12
Fix run local issue with new subnet code

JonHolman:
### Description
Fix run local issue with new subnet code.  @peoplespete what do you think about this fix?  I don't like that with this fix, if we forget to define kafkaAuthorizedSubnetIds in the deployment secret, the kafka lambda functions will have the same error as before, but as long as that is defined that issue is correct and run local will work as before without that needing to be defined.

### Related ticket(s)
CMDCT-4414

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/stacks/api.ts
+++ b/deployment/stacks/api.ts
@@ -52,7 +52,8 @@ export function createApiComponents(props: CreateApiComponentsProps) {
   Tags.of(scope).add("SERVICE", service);
 
   const vpc = ec2.Vpc.fromLookup(scope, "Vpc", { vpcName });
-  const kafkaAuthorizedSubnets = getSubnets(scope, kafkaAuthorizedSubnetIds)
+  const kafkaAuthorizedSubnets =
+    getSubnets(scope, kafkaAuthorizedSubnetIds) || vpc.privateSubnets;
 
   const kafkaSecurityGroup = new ec2.SecurityGroup(
     scope,
--- a/deployment/utils/vpc.ts
+++ b/deployment/utils/vpc.ts
@@ -1,11 +1,14 @@
 import { Construct } from "constructs";
 import { aws_ec2 as ec2 } from "aws-cdk-lib";
 
-export function getSubnets(
-  scope: Construct,
-  subnetIds: string
-): ec2.ISubnet[] {
-  return subnetIds.split(',').map(subnetId =>
-    ec2.Subnet.fromSubnetId(scope, `Subnet-${subnetId}`, subnetId)
-  )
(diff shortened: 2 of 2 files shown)


deployment/stacks/api.ts
```diff
@@ -52,7 +52,8 @@ export function createApiComponents(props: CreateApiComponentsProps) {
   Tags.of(scope).add("SERVICE", service);
```
peoplespete, line 53: I wonder if we could solve it [here](https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/master/deployment/local/prerequisites.ts) and then have a matching values in seds-defaults so that when you run local it finds your local thing as if it were the cloud thing. This is how we solved this same thing for vpnName. I'm fine with a new approach if you prefer this, more than anything I just want it to be consisten.

deployment/stacks/api.ts
```diff
@@ -52,7 +52,10 @@ export function createApiComponents(props: CreateApiComponentsProps) {
...
 
   const vpc = ec2.Vpc.fromLookup(scope, "Vpc", { vpcName });
-  const kafkaAuthorizedSubnets = getSubnets(scope, kafkaAuthorizedSubnetIds)
+  const kafkaAuthorizedSubnets = getSubnets(
+    scope,
+    kafkaAuthorizedSubnetIds ?? ""
```
peoplespete, line 57: It this still necessary? I think it will always be there. Oh is this because of the dang bootstrap thing. Ugh. That's cool.

peoplespete (dismissed):
nice thanks for reworking this, I like it!

deployment/stacks/api.ts
JonHolman: yeah, the bootstrap :(

deployment/local/prerequisites.ts
```diff
@@ -1,28 +1,44 @@
...
+    const localstackVpc = new ec2.Vpc(this, "localstackVpc", {
       ipAddresses: ec2.IpAddresses.cidr("10.0.0.0/16"),
       enableDnsSupport: true,
       enableDnsHostnames: false,
       subnetConfiguration: [],
+      vpcName: "localstack",
     });
 
-    Tags.of(mdctVpcEastDev).add("Name", "mdct-vpc-east-dev");
+    Tags.of(localstackVpc).add("Name", "localstack");
```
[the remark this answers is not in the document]
JonHolman: I agree, we probably don't need both.

peoplespete (approved):
🌮

PR #15169, opened 2025-09-08
[SEDS] Update Tealium path

rocio-desantiago:
### Description
Tealium tracking hasn't been working as expected. This PR updates the Tealium path following documentation [here](https://confluence.cms.gov/spaces/BLSTANALYT/pages/432352891/Tealium+Implementation+Documentation+for+Single-Page+Applications). Once we've confirmed with BlastX that they are seeing the expected analytics data in Adobe then we can move forward with making the same changes across all apps.

### Related ticket(s)

CMDCT-5066

---

### How to test
Once merged, will set up a meeting with BlastX to test

### DIFF
--- a/services/ui-src/index.html
+++ b/services/ui-src/index.html
@@ -33,13 +33,7 @@
       "mdctsedsval.cms.gov": "qa"
     };
     const tealiumEnv = tealiumEnvMap[window.location.hostname] || "dev";
-    const tealiumDomainMap = {
-      "prod": "tags.tiqcdn.com",
-      "qa": "tags.tiqcdn.com",
-      "dev": "tealium-tags.cms.gov",
-    }
-    const tealiumDomain = tealiumDomainMap[tealiumEnv];
-    const tealiumPath = `https://${tealiumDomain}/cms-general/${tealiumEnv}`;
+    const tealiumPath = `https://tealium-tags.cms.gov/cms-general/${tealiumEnv}`;
     document.write(`<script src="${tealiumPath}/utag.js" async><\/script>`);
     document.write(`<script src="${tealiumPath}/utag.sync.js"><\/script>`);
   </script>
--- a/services/ui-src/index.html
+++ b/services/ui-src/index.html
@@ -20,12 +20,12 @@
         Learn how to configure a non-root public URL by running `npm run build`.
       -->
     <title>CHIP Statistical Enrollment Data Reports</title>
-    <script src="/env-config.js" type="module"></script>
+    <script src="/env-config.js"></script>
     <link rel="stylesheet" type="text/css"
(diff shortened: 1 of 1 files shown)


services/ui-src/index.html
```diff
@@ -1,64 +1,65 @@
...
-  <!--
-      This HTML file is a template.
-      If you open it directly in the browser, you will see an empty page.
-
-      You can add webfonts, meta tags, or analytics to this file.
-      The build step will place the bundled scripts into the <body> tag.
-
-      To begin the development, run `npm start` or `yarn start`.
-      To create a production bundle, use `npm run build` or `yarn build`.
-    -->
```
benmartin-coforma, line 60: Yes! Begone, comment! Thank you Rocio ❤️

services/ui-src/index.html
```diff
@@ -1,64 +1,65 @@
...
+        a = `https://tealium-tags.cms.gov/cms-general/${tealiumEnv}/utag.js`;
+        b = document;
+        c = "script";
+        d = b.createElement(c);
+        d.src = a;
+        d.type = "text/javascript";
+        d.async = true;
+        a = b.getElementsByTagName(c)[0];
+        a.parentNode.insertBefore(d, a);
+      })();
```
benmartin-coforma, line 58: Not a suggestion! But I need to complain: it kills me to see all this obfuscation that really could & should just be one line, on any modern browser:

```suggestion
      document.head.append(`<script async src="https://tealium-tags.cms.gov/cms-general/${tealiumEnv}/utag.js"></script>`);
```

But I guess the folks at CMS/tealium really want to make sure the analytics can run even on ~decade-old browsers~... ~"security solutions"~... ~piles of trash~ Internet Explorer, which never supported using `append` that way. If this passes our testing, then so be it.

rocio-desantiago: Is it really just for internet explorer?

benmartin-coforma: Yep. There are a few ways to add elements to the DOM, like `.appendChild()`, `.insertBefore()`, `.append()`, and some others. One of the advantages of `.append()` is that [it supports strings](https://developer.mozilla.org/en-US/docs/Web/API/Element/append), so that you don't have to instantiate your new element ahead of time with `document.createElement()`.

And [support for `append()` started really spreading in 2016](https://developer.mozilla.org/en-US/docs/Web/API/Element/append), and has been all but guaranteed since 2018. In _real_ browsers, that is. You can see the red column for IE 11 there, which was released in 2013. Not that it would have been likely to have support for this, even if it was supported by other browsers at the time. IE was never famous for its compatibility with web standards.

benmartin-coforma (approved):
Thank you Rocio!

PR #15170, opened 2025-09-09
Convert UI tests from jest to vitest

benmartin-coforma:
### Description
This PR moves the SEDS frontend unit tests from `jest` to `vitest`.

The primary benefit is unification of build configuration. We have configured `vite` to produce a valid build of the app, which is what we actually deploy. And, if we use `jest` to test, we must _also_ configure that to produce a valid build of the app, which _hopefully_ matches what we deploy. But `vitest` can use the same configuration settings as `vite`! So duplication is reduced, and the possibility of config drift for deploy vs test is all but eliminated.

The interfaces and behaviors of both libraries are very similar; `vitest` was designed as a drop-in replacement for `jest`. The main things you'll see in this PR:
* Pretty much everything that used to say `jest.something` now says `vi.something`
* Calls to `jest.requireActual` became calls to `importOriginal`, which behaves slightly differently, in that it is async
* All of the normal testing functions (`describe`, `expect`, and so on) are imported explicitly at the top of each test file.
   + Strictly speaking this wasn't necessary; it is possible to configure vitest/typescript to import these functions implicitly everywhere. But I think the explicit imports are better, so I didn't bother fiddling with that.
* If SEDS had any tests with mock timers, you would see differences there as well. But we don't have any tests like that right now.

Using `jest` also required us to install `babel`, to transform our JSX markup to plain JS. Since `vite` ships with its own transpilation, making the switch to `vitest` also reduces our dependencies:
```
 + vitest
 + @vitest/coverage-v8
 - @babel/core
 - @babel/plugin-proposal-class-properties
 - @babel/plugin-transform-react-jsx
 - @babel/preset-env
 - babel-eslint
 - babel-jest
 - identity-obj-proxy
 - jest
 - jest-environment-jsdom
 - jest-watch-typeahead
```

### Related ticket(s)
n/a

---

### How to test
1. Observe that the frontend unit tests ran just fine on this PR
2. Observe that the site deployed as normal

### Notes
I'm never sure what to put in the notes, but this PR template says that "changed dependencies" is noteworthy. Well, this PR changes our dependencies. See above for details.

---

### Pre-review checklist

---

#### Review

#### Security


rocio-desantiago (approved):
Makes sense 😃

Filtos (approved):
Huge. Looks good!

cy-by (approved):
Worked locally, after figuring out some personal issues that `nvm use` fixed.

PR #15350, opened 2026-03-13
Remove fontawesome and replace with just icons we use in components/I…

JonHolman:
### Description
Remove fontawesome and replace with just icons we use in components/Icons.jsx

### Related ticket(s)

CMDCT-5764

---
### How to test

https://d2bfbdlo7v2twq.cloudfront.net/

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


stacy-coforma (changes requested):

stacy-coforma (dismissed):
Lastly (probably 😅) I think I'd like to see the icons move out of the `components` directory structure since they are not themselves dedicated components; perhaps we can follow HCBS's pattern here and place them under something like `src/assets/icons/`

I appreciate the changes you have made thus far!! I added one a11y note for the one spot I found that appears to be more than a decorative icon.

PR #15386, opened 2026-04-10
UI typescript

benmartin-coforma:
### Description
This PR converts all JS files in the SEDS UI to TS files. It also greatly lowers the strictness of the TS settings, so that we don't have to immediately rewrite the app.

This opens up the door for more type checking. I envision a series of future PRs, each introducing a new more specific type (for example, the shape of a record in `AuthUser`), and introducing that type wherever appropriate throughout the UI code. Eventually, we will be able to ratchet the TS settings back up, until we have strict typing throughout the app.

#### Change details:
- Various files: Quash TS complaints by adding casts. Usually to `any`, in some cases to `number` or `Record<string, number>`. This does not change runtime behavior.
- Various files: Explicitly pass `undefined` as a prop value, for required props which were omitted. This should not change runtime behavior.
  - For example, the TextFields in EditUser.tsx.
- Various test files: Distinguish between imported methods (which TS treats as their actual implementations) and mocked methods (which `vitest` additionally adorns with `.mockReturnValue()` and such).
  - For example, EditUser.test.tsx.
  - The `vi.mocked()` utility makes this pretty painless; it's an identity function (`vi.mocked(x) returns x`), but with a cast so that TS accepts it as a mock now ("look again - the function is now diamonds")
- Various test files: Add calls to `node:assert`, to convince TS that we have the type of thing we expect
  - For example, asserting that we found image elements in Footer.test.tsx, allowing us to look at `image.src` without TS complaints.
- App.tsx: The Header component does not accept a `displayHeader` prop; it always displays.
- AppRoutes.tsx: It doesn't matter what we import from "react", so long as we import it in every file.
- FormPage.tsx: Does not accept a `lastModified` prop; it pulls the modify timestamp from the zustand store.
- GenerateQuarterForms.test.tsx: An improvement! TS now has enough information to warn us that we are mocking an async function, and thus should `.mockResolvedValue()` rather than `.mockReturnValue()`
- LoaderButton.test.tsx: `isLoading` is a required property; let's not omit it.
- PrintPDF.tsx: I don't understand what this code's authors thought it was doing, but it's clear to me that the correct behavior should be "If we have state comments, display them". So that's what we're doing now.
- Profile.test.tsx: Here I made the assertion more type-agnostic ("input should have value foo" instead of "input.value should be foo"), rather than making the types involved more specific.
- Profile.tsx: This file's only change was to add `name={undefined}` to each input; however, that resulted in a bunch of new line breaks due to the formatter, which meant enough changed lines that git got confused. It therefore appears in this diff as a delete-and-add rather than a rename.
- Question.tsx, Question.test.tsx, SynthesizedGrid.test.tsx: The SynthesizedGrid component does not accept questionID or GridData props. It knows that it is only used for question 5, and it reads answer data from the zustand store.
- SummaryTab.tsx: The Question component does not accept a questionID prop. It reads the questionID from the questionData prop.
- config.ts: The `window` object does not, by default, have an `_env_` property. We add it dynamically at runtime. This TS Ignore directive is in line with our other apps.
- api.ts: This little helper type could have been an `any` to prevent complaints throughout the rest of the file (whenever we try to set body on a request). But I felt like going the extra mile and being more specific.
- store.ts: Likewise, I think I could have gotten away with an `any`, but `StoreShape` gives us a little more safety throughout the app. Someday we will revisit this, to be more specific about the shapes of questions, answers, forms, and users.

### Related ticket(s)
n/a

---
### How to test
https://d1w1s7w6ckejo2.cloudfront.net/

Does the site still load?

---
### Pre-review checklist

---
#### Review

#### Security


rocio-desantiago (approved):
All working still.
Thanks for the descriptive overview!

stacy-coforma (approved):

PR #15387, opened 2026-04-14
CMDCT-5808 destroying dependabot branches reliably with revert to see if it actually works

peoplespete:
### Description
Many times dependabot stacks aren't getting deleted

### Related ticket(s)
CMDCT-5808

---
### How to test
I'm reverting this https://github.com/Enterprise-CMCS/macpro-mdct-seds/pull/15383/changes
That way once we merge this and wait a few days dependabot should try the fix again and we should be able to see it clean up after itself this time. If that happens we'll roll this change out to core.

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/.github/workflows/destroy.yml
+++ b/.github/workflows/destroy.yml
@@ -3,25 +3,26 @@ name: Destroy
 
 on:
   delete:
-  workflow_dispatch:
-    inputs:
-      environment:
-        description: "Name of the environment to destroy:"
-        required: true
+  pull_request:
+    types:
+      - closed
 
 concurrency:
   # Same group as deploy.yml to prevent simultaneous deploy/destroy
-  group: ${{ inputs.environment || github.event.ref }}
+  group: ${{ github.event.ref }}
 
 permissions:
   actions: read
   contents: read
   id-token: write
 
+env:
+  branchName: ${{ github.event.pull_request.head.ref || github.ref_name }}
+
 jobs:
   destroy:
     name: Destroy
-    if: (inputs.environment || github.event.ref_type == 'branch') && !contains('skipci', github.event.ref)
+    if: ((github.event.ref_type == 'branch') && !contains('skipci', github.event.ref)) || (github.event.pull_request && startsWith(github.event.pull_request.head.ref, 'dependabot'))
     environment: dev
     runs-on: ubuntu-latest
     steps:
@@ -38,7 +39,7 @@ jobs:
 
       - name: Set branch_name
         run: |
-          BRANCH_NAME=$(./.github/setBranchName.ts ${{ inputs.environment || github.event.ref }})
+          BRANCH_NAME=$(./.github/setBranchName.ts ${{ env.branchName }})
           echo "branch_name=${BRANCH_NAME}" >> $GITHUB_ENV
 
       - name: Configure AWS credentials for GitHub Actions
@@ -46,6 +47,7 @@ jobs:
         with:
           aws-region: us-east-1
           role-to-assume: ${{ secrets.AWS_OIDC_ROLE_TO_ASSUME }}
+
       - name: Destroy
         run: ./run destroy --stage $branch_name --verify false
         env:
--- a/.github/workflows/destroy.yml
+++ b/.github/workflows/destroy.yml
@@ -6,10 +6,11 @@ on:
   pull_request:
     types:
       - closed
-
-concurrency:
-  # Same group as deploy.yml to prevent simultaneous deploy/destroy
-  group: ${{ github.event.ref }}
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Name of the environment to destroy:"
+        required: true
 
 permissions:
   actions: read
@@ -17,12 +18,12 @@ permissions:
   id-token: write
 
 env:
-  branchName: ${{ github.event.pull_request.head.ref || github.ref_name }}
+  branchName: ${{ inputs.environment || github.event.pull_request.head.ref || github.event.ref }}
 
 jobs:
   destroy:
     name: Destroy
-    if: ((github.event.ref_type == 'branch') && !contains('skipci', github.event.ref)) || (github.event.pull_request && startsWith(github.event.pull_request.head.ref, 'dependabot'))
+    if: inputs.environment || ((github.event.ref_type == 'branch') && !contains(github.event.ref, 'skipci') && !startsWith(github.event.ref, 'dependabot/')) || (github.event.pull_request && startsWith(github.event.pull_request.head.ref, 'dependabot/'))
     environment: dev
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/destroy.yml
```diff
@@ -3,25 +3,22 @@ name: Destroy
...
   delete:
-  workflow_dispatch:
-    inputs:
-      environment:
-        description: "Name of the environment to destroy:"
-        required: true
-
-concurrency:
-  # Same group as deploy.yml to prevent simultaneous deploy/destroy
-  group: ${{ inputs.environment || github.event.ref }}
```
peoplespete, line 14: This didn't actually match deploy.yml's so it wasn't doing anything. Since it wasn't doing anything we've already run the experiment on how big a problem not having it is...

.github/workflows/destroy.yml
```diff
@@ -3,25 +3,22 @@ name: Destroy
...
+  pull_request:
+    types:
+      - closed
```
gmrabian, line 8: will `pull_request closed` and `branch delete` cause it to race? When we close a PR we delete the branch, so will it initiate two destroys and one will fail?

peoplespete: The issue is that the branch delete hasn't been happening at all (that I've been able to find) for dependabot branches.

peoplespete: So in this new world:
if you're any branch other than dependabot and you're deleted, it triggers destroy
if you're dependabot and your PR gets closed, it triggers destroy

gmrabian: > if you're any branch other than dependabot and you're deleted, it triggers destroy

ok so I guess my question is about this case — will normal PRs double-trigger destroy?

.github/workflows/destroy.yml
```diff
@@ -3,25 +3,22 @@ name: Destroy
 
 on:
   delete:
```
JonHolman, line 5: Can we please keep `workflow_dispatch`?

package.json
```diff
@@ -43,7 +43,7 @@
   "devDependencies": {
     "@octokit/auth-action": "^6.0.1",
     "@octokit/rest": "^22.0.0",
-    "esbuild": "^0.27.7",
```
JonHolman, line 46: why downgrade?

services/ui-src/package.json
```diff
@@ -36,7 +36,7 @@
     "dotenv": "8.2.0",
     "jsdom": "^27.0.0",
     "typescript": "^5.9.3",
-    "vite": "^6.4.2",
```
JonHolman, line 39: why downgrade?

.github/workflows/destroy.yml
gmrabian: Also this will start deleting PRs that are closed but not necessarily merged. So if someone is working on something and closes the PR for some reason other than a merge, it will start destroying even if they intended to keep working on it and reopen later

.github/workflows/destroy.yml
```diff
@@ -3,25 +3,22 @@ name: Destroy
...
   id-token: write
 
+env:
+  branchName: ${{ github.event.pull_request.head.ref || github.ref_name }}
+
 jobs:
   destroy:
     name: Destroy
-    if: (inputs.environment || github.event.ref_type == 'branch') && !contains('skipci', github.event.ref)
+    if: ((github.event.ref_type == 'branch') && !contains('skipci', github.event.ref)) || (github.event.pull_request && startsWith(github.event.pull_request.head.ref, 'dependabot'))
```
JonHolman: we want to get rid of skipci, right?  I mean I like having it, but I thought we removed it.

JonHolman: We should add to the delete path `&& !startsWith(github.event.ref, 'dependabot/')` 
If we keep skipci, it should be `contains(github.event.ref, 'skipci')`

JonHolman: Also, `branchName` should use `github.event.ref` for delete events instead of falling back to `github.ref_name`.

.github/workflows/destroy.yml
peoplespete: Ah ok.
This IF is what keeps these things separate:
If it's happening because of a deleted branch the first half will matter
If it's happening because of a closed PR the second half will matter

That is what allows us to trigger on both but not have to worry about the regular branches being affected by the new PR close trigger.

    if: ((github.event.ref_type == 'branch') && !contains('skipci', github.event.ref)) || (github.event.pull_request && startsWith(github.event.pull_request.head.ref, 'dependabot'))

peoplespete: The concern about a race condition would only apply to dependabot branches:

what if it:
-removes the branch and triggers
AND
-PR closes

My response is that I wish this were a problem but I've found no evidence that the first part of that race is ever happening.

.github/workflows/destroy.yml
peoplespete: They are different things afaik
github.ref_name == 'main'
github.event.ref == 'refs/heads/main'

peoplespete: skipci still lives on

peoplespete: If I added this part to the on delete:
&& !startsWith(github.event.ref, 'dependabot/')
It won't do anything
I wish it did something in which case the entire ticket wouldn't need to be worked.
That's my understanding. I can add it for good measure if you really want it.

services/ui-src/package.json
peoplespete: So that dependabot will make a new PR so that we can see if this new destroy triggering works. I just reverted the last successful PR from dependabot.
https://github.com/Enterprise-CMCS/macpro-mdct-seds/pull/15383/changes

.github/workflows/destroy.yml
JonHolman: For `delete` events, `github.event.ref` is the short branch name. `github.ref_name` doesn’t give us the deleted branch there, so branchName should use `github.event.pull_request.head.ref || github.event.ref` (and only fall back to `github.ref_name` for workflow_dispatch if we keep it).

JonHolman: I would like `&& !startsWith(github.event.ref, 'dependabot/')` on the delete path as a safety net.

.github/workflows/destroy.yml
peoplespete: Ok added extra logic so that dependabot won't trigger the branch deletion runs just in case.

.github/workflows/destroy.yml
peoplespete: Just added back

.github/workflows/destroy.yml
peoplespete: Ok should be all up to date now

JonHolman (approved):

PR #15392, opened 2026-04-22
Address various security warnings

benmartin-coforma:
### Description
This PR "fixes" the 19 currently-outstanding security warnings on this repo.

It is my belief that 0 of these warnings represented actual vulnerabilities. 
- 3 were vulnerabilities when exposing Vite's dev server to the network. We don't do that.
- 5 (form-data, protobufjs, request, qs\*2) were exposed by processing untrusted data, but were only called from within our end-to-end tests - during which all data is trusted.
- 7 (minimatch\*3, picomatch\*2, brace-expansion\*2) were theoretical DoS attacks, potentially consuming resources when evaluating inefficient code. But these were called only during development (build / test / deploy), so again: input is trusted, and our servers are safe.
- 1 was another theoretical DoS attack, via underscore, via jsonpath. Again, I believe this was unexploitable in practice, given our limited use of jsonpath and our data validation.
- 1 was a flaw allowing external communication with the esbuild development server. We don't use this feature.
- 1 was a flaw in the YAML parser used by aws-cdk-lib. We only use aws-cdk-lib to _generate_ YAML.
- 1 was a flaw in the XML parser used by various `@aws-sdk` packages. Our app doesn't interact with XML in any way that I'm aware of, again making this unexploitable in practice.

Mostly I addressed these by upgrading package versions. However, I also disabled our end-to-end test suite entirely. It currently only runs one """test""", which actually just loads the google home page. In other words, it doesn't test our site at all. I believe that disabling the test project (by renaming its `package.json` file, and removing all other references to it) is a better use of our time than to try to upgrade its various dependencies. We are dismounting the dependency-upgrade treadmill.

The only code changes related to Vite/Vitest upgrades. The new vitest version broke some of our stranger mocks, and the new vite version obsoleted a plugin. We _love_ to see a transient dependency removed.

### Related ticket(s)
n/a

---
### How to test
https://d2e3398q1mumjr.cloudfront.net/

* Ensure the site still builds, passes unit tests, deploys, and functions.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


benmartin-coforma:
I'm not sure how this PR affects test coverage at all? But -0.17% is just noise regardless. We will override this to merge.

stacy-coforma (approved):
I'm behind this. I think we'll see Playwright replacing Cypress anyway in the long run, and I'm on the fence over even keeping the Cypress code around but removing the job in CI is a good move.

benmartin-coforma:
> I'm behind this. I think we'll see Playwright replacing Cypress anyway in the long run, and I'm on the fence over even keeping the Cypress code around but removing the job in CI is a good move.

You make a good point. I've updated the upcoming "Remove Cypress from SEDS" story (CMDCT-5795) to clarify that the requirements are now minimal:

> * Read each test, to see if it would test anything useful (if it were to actually run).
>    * If so, and if that test isn't already noted in tests/manual/FullRegression.md, add it there.
> * Delete the tests/cypress directory

rocio-desantiago (approved):
Looks good 👍

PR #15413, opened 2026-05-15
Storage helpers

benmartin-coforma:
### Description
Separate business logic from database interaction, by moving database queries into app-api/storage.

New functions:
   - `formAnswers/queryAnswersByEntry()`
   - `formAnswers/queryAnswersByForm()`
   - `formAnswers/updateAnswer()`
   - `formQuestions/scanQuestionsByYearAndForm()`
   - `formTemplates/scanTemplateYears()`
   - `stateForms/scanFormsByState()`
   - `stateForms/scanFormsByStateAndQuarter()`
   - `stateForms/scanFormsWithTotals()`
   - `stateForms/getStateForm()`
   - `stateForms/updateEnrollmentCounts()`
   - `stateForms/updateComment()`
   - `stateForms/updateCommentAndStatus()`
   - `users/scanAllUsers()`
   - `users/getUser()`
   - `users/recordLogin()`

The loose convention for these helper functions is to name them after the underlying DynamoDB operation. For example, if we ultimately send a `ScanCommand` then the function should be named `scanFoo()`. Another loose convention is that, if the function takes many parameters which are fields of the database object, it should accept an instance of the database object (with at least those fields present, specified using the `Pick<type, keys>` utility https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys).

The biggest change was to handlers/forms/updateForm.ts. Its primary purpose is to update the answers for a form; that part was a clean swap to the new `updateAnswer()` function. But it also updates the form metadata: the last modify date, the form-level comment field, the status, and - _if the status changed_ - the status modify date. Previously that conditionality was handled by changing the parameters to the `UpdateCommand`. Now, it's handled by calling one of two separate update functions. (The other change in this file was to remove the unused `questionResult` variable.)

When I changed a handler file to remove direct DynamoDB interaction, I also changed its tests to be less direct. For example, handlers/admin/forms/getForm.ts used to call `dynamodb.get()`, and its tests mocked the responses to `GetCommand`. Now, instead it calls `getStateForm()` in storage/stateForms.ts, and its tests mock that function.

I also reworked the tests for the storage code. Previously, these tests mocked DynamoDB interactions using the `aws-sdk-client-mock` package. Now, they use more direct mocking: `vitest` built-in mocks, via dynamo-mocking.ts.

For example, this test
```ts
import { getFoo } from "../storage/foo.ts";
import { DynamoDBDocumentClient, GetCommand } from "@aws-sdk/lib-dynamodb";
import { mockClient } from "aws-sdk-client-mock";
const mockDynamo = mockClient(DynamoDBDocumentClient);
const mockGet = vi.fn();
mockDynamo.on(GetCommand).callsFake(mockGet);

// example use in an actual test
it("should work", async () => {
  mockGet.mockResolvedValueOnce({ Item: mockData });
  expect(await getFoo("my-foo-id")).toEqual(mockData);
  expect(mockGet).toHaveBeenCalledWith(
    { TableName: "foo-table", Key: { id: "my-foo-id" } },
    expect.any(Function)
  );
});
```
would now look like
```ts
import { mockGet } from "../libs/dynamo-mocking.ts";
import { getFoo } from "../storage/foo.ts"; // imported AFTER dynamo-mocking.ts

// test code is unchanged, except we no longer expect mockGet to receive a callback
it("should work", async () => {
  mockGet.mockResolvedValueOnce({ Item: mockData });
  expect(await getFoo("my-foo-id")).toEqual(mockData);
  expect(mockGet).toHaveBeenCalledWith({ TableName: "foo-table", Key: { id: "my-foo-id" } });
});
```
This obviously makes the test files briefer - albeit a little more magical, since the mocks are now defined elsewhere. The bigger gain is that if we switch to use this approach exclusively, we can completely eliminate our dependency on `aws-sdk-client-mock` (and thus also `sinon`, `just-extend`, `nise`, `path-to-regexp`, and `type-detect`).

This PR doesn't go that far; it doesn't touch all of the handler files, so some of their tests still pull in `aws-sdk-client-mock`. But this PR does pave the way for that eventual future.

### Related ticket(s)
CMDCT-5130

---
### How to test
https://d3gwlb7kou0rpk.cloudfront.net/

There should be no behavioral changes. A light regression test will be sufficient.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/app-api/storage/stateForms.ts
```diff
@@ -87,6 +119,104 @@ export const scanFormsByQuarterAndStatus = async (
...
+      ":f2": "64.21E",
+    },
+    ConsistentRead: true,
+  });
+  return response.Items as StateForm[];
+};
+
+export const getStateForm = async (state_form: StateForm["state_form"]) => {
+  const response = await dynamoDb.get({
+    TableName: process.env.StateFormsTable,
```
stacy-coforma:
```suggestion
    TableName,
```
right? I think this is still set to `process.env.StateFormsTable` from earlier?

services/app-api/storage/stateForms.ts
```diff
@@ -87,6 +119,104 @@ export const scanFormsByQuarterAndStatus = async (
...
+};
+
+export const updateEnrollmentCounts = async (
+  data: Pick<
+    StateForm,
+    "state_form" | "last_modified" | "last_modified_by" | "enrollmentCounts"
+  >
+) => {
+  await dynamoDb.update({
+    TableName: process.env.StateFormsTable,
```
stacy-coforma:
```suggestion
    TableName,
```

iff the one above is also correct 😅

services/app-api/handlers/notification/uncertified.ts
```diff
@@ -51,28 +51,8 @@ export const main = handler(emptyParser, async (request) => {
...
-  payload.map((userInfo) => {
-    if (userInfo.email) {
-      businessOwnersEmails.push(userInfo.email);
-    }
-  });
-  return businessOwnersEmails;
-}
-
 async function unCetifiedTemplate(payload: RequestBody) {
-  const sendToEmail = await getBusinessUsersEmail();
```
stacy-coforma, line 75: excellent. `unCetified` really bothers me but you did not introduce this 😅

stacy-coforma (commented):
leaving a note to confirm I have reviewed, but I want to also test before I approve and I'm out of steam for that tonight. Will get to that in the morning!

services/app-api/storage/stateForms.ts
benmartin-coforma: Yes, nice catch. Updated these and 2 others; that should be all of them.

services/app-api/handlers/notification/uncertified.ts
benmartin-coforma: It's sort of beautiful in a way, isn't it? That's my emotional support legacy typo.

stacy-coforma (dismissed):
Tested with a variety of users, poked around in several random areas and everything seems to be saving/updating/displaying as expected.  Nice work!

stacy-coforma (approved):

sopspops (approved):
Tested and it works good!!

PR #15415, opened 2026-05-19
Update getCurrentUser endpoint

sopspops:
### Description

* update getCurrentUser to determineCurrentUser
* Change the method to POST from GET
* Switch 500 status to 401

### Related ticket(s)

CMDCT-5798

---
### How to test

* Test this by opening the browser console before logging in to SEDS, you will see an error in the console with the correct error

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/app-api/libs/handler-lib.ts
```diff
@@ -43,6 +46,10 @@ export default function handler<TParams>(
 
       return await lambda({ user, body, parameters });
     } catch (error) {
+      if (error instanceof UnauthenticatedError) {
```
stacy-coforma: Just peeping the deploy error - not sure if this came from a previous iteration of the changes?  I don't see `UnauthenticatedError` in `./authorization.ts` so I'm wondering if either that change is missing or this one needs to be removed?

sopspops: oooope I was refactoring it and pushed it without cleaning up entirely!

services/app-api/libs/handler-lib.ts
```diff
@@ -55,11 +55,16 @@ export default function handler<TParams>(
...
+    const userFromToken = getUserDetailsFromEvent(event);
+    if (event.path === "/determineCurrentUser") {
+      // determineCurrentUser creates AuthUser records, so they needn't already exist.
+      return userFromToken as AuthUser;
+    } else {
+      return await scanForUserWithSub(userFromToken.usernameSub);
+    }
+  } catch {
+    // Auth parsing failed - return 401
+    return null;
```
benmartin-coforma: This is a good idea: if we receive a request with an invalid token then 401 is the ideal HTTP response. Since the way we would notice and invalid token is by `getUserDetailsFromEvent()` throwing an error, this try-catch does the job.

But there are three problems:
1. The try block also includes `scanForUserWithSub()`. If that throws, then the problem is not with the request but with our backend, and a 500 is appropriate again.
2. The catch block swallows the error; it is neither logged nor propagated back to the calling method. So if we do get an error, we won't be able to find out what it was or where it was thrown. 
3. We prefer `undefined` over `null` as the "missing value" value ;)

Problem 2 is the reason that there are almost no `catch` blocks in the entire SEDS backend. Any error thrown from any lambda (with a few exceptions in unmaintained or non-http lambdas) will bubble all the way back up to this file, to be logged and turned into a 500.

benmartin-coforma (commented):
This is very nice work - all of the `determineCurrentUser` changes look right to me.

For my comment below, you could either address those problems by reworking `determineUser()` in handler-lib.ts, or by moving the try-catch down a level into `getUserDetailsFromEvent()` in authorization.ts, and having that method return `undefined` (and do its own logging of the reason) instead of throwing errors.

Which of those approaches makes more sense to you? I could see arguments for either.

services/app-api/libs/handler-lib.ts
sopspops: Thank you Ben!! ohhh that totally makes sense!! I'll move the try-catch block down a level since I feel like it's simpler hehe

benmartin-coforma (approved):
Marvelous

PR #15420, opened 2026-05-28
CMDCT-5764 Replace FontAwesome with SVG icons.

lgilders-coforma:
### Description

Remove FontAwesome and replace with SVG icons.

### Related ticket(s)

[CMDCT-5764](https://jiraent.cms.gov/browse/CMDCT-5764)

---
### How to test

Review the existing application in the dev or val environments and compare with the PR version.

The affected icons requires both the StateUser and AdminUser credentials.

AdminUser:
AdminUser Edit Template Screen
AdminUser Edit User Screen
AdminUser View Forms Arizona Screen
AdminUser View Forms GRE Screen

StateUser:
GRE Print Screen
GRE Screen
GRE Screen 2
GRE Screen 3
Login Screen
Login Screen 2
Reports Screen
Select State Screen

### Notes

The new assets folder only contains copies of the implemented SVG icons for visual reference in the IDE. The Icons.tsx file is the implementation of the SVGs. See included READ.ME for details.

There are 2 version of the FilePdf SVG - one for when it doesn't need to be identified by the screenreader and one when it does. Open to alternate implementation ideas for this use case.

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/assets/icons/access/icon_sign_in_alt.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512" width="..." height="..." aria-hidden="true" focusable="false">
```
cy-by: I would define a sane, small value for `width` and `height` so that if CSS doesn't load, or the page takes a long time to load, the SVG's are not `512px` in the mean time. A sane value would be something like `24px`. Plus these attribute get overridden once CSS loads, so they're just temporary.

services/ui-src/src/assets/icons/arrows/icon_arrow_left.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512" width="..." height="..." aria-hidden="true" focusable="false">
```
cy-by: I'm open to discussion, but I wouldn't include `aria-hidden` or `focusable` (those `focusable` is only for IE 🪦) on the base SVG and leave those attributes to be added at implementation time, since someone needs context to know if an icon should be hidden or not. What do you think?

services/ui-src/src/assets/READ.ME
```diff
@@ -0,0 +1,14 @@
+## To add a new icon to the project:
+  1. Create a new folder, if needed
+  2. Create a new file
+  3. Copy/paste in the SVG code
+  4. If the SVG does not need to be available to the screen reader, ensure it has the following properties: aria-hidden="true" focusable="false"
+  If the SVG does need to be available to the screen reader, ensure it has the following properties: aria-hidden="false" focusable="true"
```
cy-by: See comment above about these props

services/ui-src/src/components/FormFooter/FormFooter.tsx
```diff
@@ -17,7 +16,7 @@ const FormFooter = ({ state, year, quarter }) => {
     <div data-testid="FormFooter" className="form-footer">
       <div>
         <Link to={quarterPath}>
-          <FontAwesomeIcon icon={faArrowLeft} /> Back to {`Q${quarter} ${year}`}
+          <ArrowLeft /> Back to {`Q${quarter} ${year}`}
```
cy-by: For example, this `ArrowLeft` icon does need to be hidden because it doesn't add useful information for this link text. But maybe there's a different use case that needs it to be visible somewhere else in the app (probably not right now). So I'd prefer to add `aria-hidden="true"` inline here, instead of having it in the `ArrowLeft` file.

services/ui-src/src/components/Icons/Icons.tsx
```diff
@@ -0,0 +1,133 @@
...
+    width="..."
+    height="..."
+    aria-hidden="true"
+    focusable="false"
+  >
+    <path
+      fill="currentColor"
+      d="M416 448h-84c-6.6 0-12-5.4-12-12v-40c0-6.6 5.4-12 12-12h84c17.7 0 32-14.3 32-32V160c0-17.7-14.3-32-32-32h-84c-6.6 0-12-5.4-12-12V76c0-6.6 5.4-12 12-12h84c53 0 96 43 96 96v192c0 53-43 96-96 96zm-47-201L201 79c-15-15-41-4.5-41 17v96H24c-13.3 0-24 10.7-24 24v96c0 13.3 10.7 24 24 24h136v96c0 21.5 26 32 41 17l168-168c9.3-9.4 9.3-24.6 0-34z"
+    />
+  </svg>
```
cy-by: I see that you define SVG's in two spots:
1. individual `.svg` files in `src/assets/icons/`
2. individual exports here in `Icons.tsx`

What benefits are there in having two places where SVG code could change? Couldn't this file import the SVG/Path data from those files?

cy-by (changes requested):
Request changes/your thoughts on some things.

services/ui-src/src/components/Print/PrintPDF.tsx
```diff
@@ -77,7 +76,7 @@ const PrintPDF = () => {
 
           <Button variation="solid" onClick={(e) => handlePrint(e)}>
             Print / PDF
-            <FontAwesomeIcon icon={faPrint} />
+            <Print />
```
rocio-desantiago: Just a thought: would it be more reader friendly to render `<Print />` such as `<PrintIcon />`?

services/ui-src/src/components/Quarterly/Quarterly.tsx
```diff
@@ -94,7 +93,7 @@ const Quarterly = () => {
                             form
                           )}`}
                         >
-                          <FontAwesomeIcon icon={faFilePdf} />
+                          <FilePdfVisible />
```
rocio-desantiago: `<FilePdfVisibleIcon />` ?

services/ui-src/src/components/StateSelector/StateSelector.tsx
```diff
@@ -89,7 +88,7 @@ const StateSelector = () => {
             }}
           >
             Update User
-            <FontAwesomeIcon icon={faUserCheck} />
+            <UserCheck />
```
rocio-desantiago: `<UserCheckIcon />` what do we think about this naming pattern?

rocio-desantiago (dismissed):
Curious about naming convention.
i.e, `<PrintIcon />`

services/ui-src/src/components/Icons/Icons.tsx
```diff
@@ -0,0 +1,133 @@
+import { SVGProps } from "react";
+
+// Ensure SVG contains only valid HTML data
+
+export const SignInAlt = (_props: SVGProps<SVGSVGElement>) => (
```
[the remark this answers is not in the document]
lgilders-coforma: Shifted each Icon to its own component file. Added aria attribute via props. Added default width and height.

services/ui-src/src/components/Icons/Icons.tsx
```diff
@@ -0,0 +1,133 @@
...
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 512 512"
+    width="..."
+    height="..."
```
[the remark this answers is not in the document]
lgilders-coforma: Shifted each Icon to its own component file. Added aria attribute via props. Added default width and height.

services/ui-src/src/components/Icons/Icons.tsx
lgilders-coforma: Shifted each Icon to its own component file. Removed duplicated assets folder and contents.

services/ui-src/src/components/FormFooter/FormFooter.tsx
lgilders-coforma: Added aria property to props.

services/ui-src/src/additional-scss/_globals.scss
```diff
@@ -46,6 +46,32 @@
   width: 46rem;
 }
 
+//-----------------------
+//  Icons
+//-----------------------
+
+svg {
+  display: block;
```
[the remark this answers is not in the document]
lgilders-coforma: Rescoped to icon className.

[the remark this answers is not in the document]
lgilders-coforma: Shifted Footer CSS to existing .form-footer class.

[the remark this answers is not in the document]
lgilders-coforma: Removed extraneous button class.

services/ui-src/src/assets/READ.ME
```diff
@@ -0,0 +1,14 @@
...
+  Also include a <title></title> and <description</description> for the SVG
+  NOTE: If both versions are needed, this will be handled during the implementation step.
+  5. Open the Icons.tsx file
+  6. Copy/paste an existing implementation for an icon
+  7. Update the name to the new icon's name
+  8. Ensure the implementation svg has the following properties for accessibility: width="..." height="..."
```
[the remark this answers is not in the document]
lgilders-coforma: Removed as extraneous now.

services/ui-src/src/assets/READ.ME
```diff
@@ -0,0 +1,14 @@
+## To add a new icon to the project:
+  1. Create a new folder, if needed
+  2. Create a new file
+  3. Copy/paste in the SVG code
+  4. If the SVG does not need to be available to the screen reader, ensure it has the following properties: aria-hidden="true" focusable="false"
+  If the SVG does need to be available to the screen reader, ensure it has the following properties: aria-hidden="false" focusable="true"
+  Also include a <title></title> and <description</description> for the SVG
```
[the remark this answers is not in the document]
lgilders-coforma: Removed as extraneous now.

services/ui-src/src/assets/READ.ME
lgilders-coforma: Moved aria properties to props for each icon component.

services/ui-src/src/components/Print/PrintPDF.tsx
lgilders-coforma: Updated naming convention for icons.

services/ui-src/src/components/StateSelector/StateSelector.tsx
lgilders-coforma: Updated naming convention for icons.

services/ui-src/src/components/Quarterly/Quarterly.tsx
lgilders-coforma: Updated naming convention for icons.

services/ui-src/src/assets/icons/access/icon_sign_in_alt.svg
lgilders-coforma: Added default values for width and height for loading.

services/ui-src/src/assets/icons/arrows/icon_arrow_left.svg
lgilders-coforma: Removed aria-hidden and focusable from icons.

services/ui-src/src/components/Icons/Icons.tsx
```diff
@@ -0,0 +1,133 @@
...
+export const FilePdfVisible = (_props: SVGProps<SVGSVGElement>) => (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 384 512"
+    width="..."
+    height="..."
+    aria-hidden="false"
+    focusable="true"
+  >
+    <title id="pdf-icon">File PDF Icon</title>
```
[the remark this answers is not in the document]
lgilders-coforma: Added ability to use generated IDs for title and desc.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+    >
+      <title id={titleId ?? `${generatedTitleId}`}>Arrow Left Icon</title>
+      <desc id={descId ?? `${generatedDescId}`}>
+        Activate this link to return to the previous page.
+      </desc>
+      <path
+        fill="currentColor"
+        d="M257.5 445.1l-22.2 22.2c-9.4 9.4-24.6 9.4-33.9 0L7 273c-9.4-9.4-9.4-24.6 0-33.9L201.4 44.7c9.4-9.4 24.6-9.4 33.9 0l22.2 22.2c9.5 9.5 9.3 25-.4 34.3L136.6 216H424c13.3 0 24 10.7 24 24v32c0 13.3-10.7 24-24 24H136.6l120.5 114.8c9.8 9.3 10 24.8.4 34.3z"
+      />
+    </svg>
```
stacy-coforma: On the previous PR (not yours) I had declared a preference for the SVGs living as separate files, but I'm understanding from this that having them inline here allows the props and fill color to be set dynamically, which wouldn't be possible if they were separate SVGs?  Trying to validate my understanding 👀

lgilders-coforma: Correct - 
Try #1 with `` returns the svg as a data content and cannot be styled.
Try #2 with a single `Icons.tsx` impl meant we needed to duplicate icons in cases where it needed the accessibility text hidden vs. visible
Try #3 allows for 1 version of the svg with added props for the `title` and `desc` for accessibility and for the use of `aria-hidden `values added per use of the component.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
+// Font Awesome Free v7.2.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2026 Fonticons, Inc.
```
rocio-desantiago: Are we still using Font Awesome or is this comment lingering?

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React, { useId } from "react";
+
+interface ArrowLeftIconProps extends React.SVGProps<SVGSVGElement> {
+  titleId?: string;
+  descId?: string;
+}
+
+const defaultProps: Partial<ArrowLeftIconProps> = {
+  titleId: "arrow-left-title",
+  descId: "arrow-left-desc",
```
rocio-desantiago: For my own knowledge/understanding, is `descId` short for `description ID`?

rocio-desantiago (commented):
We should be good to remove this comment, right?
// Font Awesome Free v7.2.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2026 Fonticons, Inc.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
stacy-coforma: We took the icons from FontAwesome, so even though we are not using their components we should still leave the image credits in there.  I think they have something along those lines in their terms if you look on their licensing details.

rocio-desantiago: oooohhhh very interesting!

rocio-desantiago (dismissed):
Code looks clean 👍

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Correct - the html tag for 'description' got shortened to 'desc' at some point.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
benmartin-coforma: Nitpick: I believe these icons are FA v5, not v7.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+      </desc>
+      <path
+        fill="currentColor"
+        d="M257.5 445.1l-22.2 22.2c-9.4 9.4-24.6 9.4-33.9 0L7 273c-9.4-9.4-9.4-24.6 0-33.9L201.4 44.7c9.4-9.4 24.6-9.4 33.9 0l22.2 22.2c9.5 9.5 9.3 25-.4 34.3L136.6 216H424c13.3 0 24 10.7 24 24v32c0 13.3-10.7 24-24 24H136.6l120.5 114.8c9.8 9.3 10 24.8.4 34.3z"
+      />
+    </svg>
+  );
+};
+
+ArrowLeftIcon.defaultProps = defaultProps;
```
benmartin-coforma: Nitpick: Default props are deprecated, or at least soon-to-be deprecated. Instead of the custom mechanism from early versions of React:
```tsx
const MyComponent = ({ foo }) => { ... };
MyComponent.defaultProps = { foo: "bar" };
```
We should use JS's built-in default function arguments:
```tsx
const MyComponent = ({ foo = "bar" }) => { ... };
```

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export default ArrowLeftIcon;
```
benmartin-coforma: Nitpick: If the `defaultProps` line is removed, it will be possible (and, in my opinion, preferable) to export the component on the same line it's declared. And I usually like named exports, although that's largely an aesthetic preference. That is, instead of
```ts
const Foo = ...;
export default Foo;
```
I'd like to see
```ts
export const Foo = ...;
```
I believe you'll find SEDS has a mix of named & default exports, but that's only because I haven't gotten around to changing them all yet. In our newer apps, I believe you'll find named exports much more often.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+      viewBox="0 0 448 512"
+      width={24}
+      height={24}
+      className="icon arrow-left-icon"
+      aria-labelledby={
+        titleId && descId
+          ? `${titleId} ${descId}`
+          : `${generatedTitleId} ${generatedDescId}`
+      }
+      {...props}
```
benmartin-coforma: Nitpick: Spreading passthrough props is a practice you'll find recommended a lot of places, but I think it's very rarely the right move in application development.

If we were writing components for a library that's meant to be widely-used, then passthrough props would allow consumers of this code the freedom to customize the underlying HTML elements as they see fit, which would be a good thing. But we're writing components for an application with a finite scope; the only consumers is us.

For example, if we had a page in this app where we wanted to specify the `preserveAspectRatio` attribute of the underlying SVG, the passthrough props would allow us to do that from that page, without touching this code. But we could also do that in a world without passthrough props, by adding `preserveAspectRatio` to this code, as an explicit prop. It's not _zero_ cost for us to make that code change, but it is a very low cost.

The downside of passthrough props is that they make the behavior of this element more difficult to analyze. Since the `{...props}` spread happens after the other props are specified, the width/className/etc may all be overridden. That means, in order to know how this component behaves, I must not only read this component, but also go find every reference to this component, to learn which props are overridden & how. Re-ordering the props here can lessen this problem by preventing overrides, but not entirely remove it, since we still can't know (without chasing down all references) the full list of provided props.

I find that keeping as much behavior as possible fully encapsulated within a small scope (in this case, a single file) is a big win for maintenance effort.

In short, instead of the catchall passthroughs:
```tsx
type MyProps = OtherProps & { foo: string };
export const MyComponent = ({ foo, ...props }: MyProps) => {
  return (
    <OtherComponent foo={foo} bar="baz" {...props} />
  );
}
// Consuming code
<MyComponent foo="bar" quux={42} />
```
I would rather see an explicit property list, pretty much every time:
```tsx
type MyProps = { foo: string; quux: number };
export const MyComponent = ({ foo, quux }: MyProps) => {
  return (
    <OtherComponent foo={foo} quux={quux} bar="baz" />
  );
}
// (Consuming code unchanged)
```

For this component, that would mean its only prop would be `aria-hidden`, since I believe that's the only one used in consuming code throughout the app.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
+// Font Awesome Free v7.2.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2026 Fonticons, Inc.
+import React, { useId } from "react";
+
+interface ArrowLeftIconProps extends React.SVGProps<SVGSVGElement> {
+  titleId?: string;
+  descId?: string;
+}
```
benmartin-coforma: Nitpick: Along the same lines as my novel about passthrough props, I would also prefer to avoid optional arguments. Although this is a much looser rule of thumb for me.

We _may_ be able to make these required, if and only if the component is not aria-hidden. That might look like this:
```tsx
type ArrowLeftIconProps =
  { "aria-hidden": true as const } |
  { "aria-hidden": false as const; titleId: string; descId: string };
```

But we could also just remove them, since they didn't turn out to be necessary for any pages in the app.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const defaultProps: Partial<ArrowLeftIconProps> = {
+  titleId: "arrow-left-title",
+  descId: "arrow-left-desc",
+};
+
+const ArrowLeftIcon = ({ titleId, descId, ...props }: ArrowLeftIconProps) => {
+  const generatedTitleId = useId();
+  const generatedDescId = useId();
```
benmartin-coforma: Comment: The `useId` hook is pretty nifty for this purpose! I don't remember ever seeing it before this PR. My best solution otherwise would have been to generate a random ID manually, and hope for the best.

Often I will recommend against using lesser-known features of React (or any framework). My logic is that if I didn't already know about it, the maintenance dev frantically trying to troubleshoot a production fire five years from now won't know about it either, and I want to minimize the amount they have to learn.

But in this case, it is a very neat solution, and nearly self-explanatory. So long as we have this problem of needing unique IDs within a possibly-repeated component, I approve of this solution. (Foreshadowing is a literary technique wherein future events are...)

services/ui-src/src/components/Quarterly/Quarterly.tsx
```diff
@@ -94,7 +93,7 @@ const Quarterly = () => {
                             form
                           )}`}
                         >
-                          <FontAwesomeIcon icon={faFilePdf} />
+                          <FilePdfIcon aria-hidden="false" />
```
benmartin-coforma: Suggestion: We could simplify the underlying aria implementation (with its contained elements and generated IDs, and its identical text from row to row in this table), by adding screen-reader-only text to the button. That would make the icon redundant, and none of the icons in the app would have or need aria labelling.

That might look something like this:
```tsx
<button ...>
  <span class="sr-only">{form.form} printable view</span>
  <FilePdfIcon aria-hidden="true" />
</button>
```
With a new block in the global styles file, something like:
```css
.sr-only {
   position: absolute;
   left: -999em;
}
```

_But_ I would definitely want to hear @cy-by 's take before going ahead with this. I know the `.sr-only` class is not unprecedented (screenshot below), but Cypress might prefer one approach over the other, and their opinions are always correct.

Image

services/ui-src/src/components/EditUser/EditUser.tsx
```diff
@@ -165,7 +164,7 @@ const EditUser = () => {
             onClick={handleUpdateClick}
           >
             Update User
-            <FontAwesomeIcon icon={faUserCheck} />
+            <UserCheckIcon aria-hidden="true" />
```
benmartin-coforma: Big suggestion: I would still very much like to see individual .svg files. I know there's been a lot of discussion and back-and-forth on this, and the shortcomings of various approaches. The approach I've found to work, which allows dynamic styling without custom JSX for each icon, is the [approach recommended by FA themselves](https://docs-v5.fontawesome.com/web/advanced/svg-sprites):
1. Download the SVG files directly from FA
   * For example, from https://fontawesome.com/v5/icons/classic/solid/sign-in-alt
2. Store them in the ui-src/public directory
   * Probably in the /img subdirectory
   * This makes them available to fetch at runtime, via a separate request from the browser
3. Edit each one to add an explicit width and height
   * Probably 16, since that's the size this app usually uses
   * Grit your teeth and try to ignore how oxfmt mangles the comment
4. Within the markup for an individual page, create a new `<svg>` element
5. But instead of embedding a `<path>`, link out to the FA file.

That would make this code look like this:
```suggestion
            <svg className="icon" aria-hidden="true">
              <use xlinkHref="/public/img/user-check-solid.svg"></use>
            </svg>
```

This will _mostly_ work, give or take some style tweaking. See my next comment, on _globals.scss

services/ui-src/src/additional-scss/_globals.scss
benmartin-coforma, line 1: Suggestion: I believe these styles will work unchanged for all of the icons contained within `<button>` elements, which is all but two of them. The Quarterly.tsx PDF icon and the FormFooter.tsx left arrow icon are inside `<a>` elements, and they'll look a little funky:

Image Image

This is because we have, at this point, two SVGs: The inner one with the actual path, which we've sized at 16x16, and the outer one acting as a container, which takes a default size of 150x300.

The default sizing, we can cancel out by setting `width: 1em; height: 1em;` within the `.icon` block. I must admit I am not a CSS expert, so I'm not sure what behavior we're losing by changing that from `width: 100%; height: auto;`, but it seems fine afaict?

That will leave the left arrow in the form footer misaligned its text. Again, I'm not an expert, but I'd probably fix that just by budging it over with with `a > .icon { .margin-right: 0.5em; margin-bottom: 0.15em; }`

It will also leave the PDF icon in quarterly table too small. My first impulse would be to fix this by adding `className="icon large"` in the markup, and styling `a > .icon.large { transform: scale(2.75); transform-origin: top left; margin: auto; }`

I _think_ that will leave the whole app looking pretty?

benmartin-coforma, line 177: One more tiny request: Could you move this rule to the `button > .icon` style you defined above, please? For consolidation purposes.

benmartin-coforma (dismissed):
I really appreciate all the work you're putting in here. This review is a bunch of informational nitpicks (hopefully none of which conflict with requests you've already fielded), followed by a big request that involves deleting all the files I had nitpicks about.

I know I'm asking a lot here, especially given how easy this story may have sounded at first. I'll be happy to pitch in and/or pair on this, if it feels like too much.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Sadly, there's one place where it's an issue - the list of reports with the File PDF icon and no explanation text. Each icon need to be visible to aria, although in this case, they would all have the same title and description. So I'm not confident if it's an applicable use case. 🤔

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Used this for that 1 use case of the File PDF icons for each report in the list. Again, each would have the same title and description, so unsure if we consider it a valid use case. 🤔

services/ui-src/src/components/Quarterly/Quarterly.tsx
lgilders-coforma: Still stuck on the 1 use case of each report in the list having just the File PDF icon and no explanation text for its action. 😭

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
cy-by: I tend to recommend `useId` because it's easier than handcrafting a custom id that thing sometimes might be unique enough (this has happened twice since my tenor on MDCT). Other reasons I like it:
- It's baked into React and not a third party
- It's [specifically for accessibility](https://react.dev/reference/react/useId#:~:text=generating%20unique%20IDs%20that%20can%20be%20passed%20to%20accessibility%20attributes), so the more we use it instead of hand crafted `id`'s, the more it will be associated with solving accessibility things in our apps.

I also agree with the use of it here in this context.

Just my 2 cents.

services/ui-src/src/components/Quarterly/Quarterly.tsx
cy-by: ## Using HTML hidden text

@benmartin-coforma, your suggestion is the top tier. It is preferred to have hidden (via CSS) HTML text over SVG labels. Now it's not wrong to use SVG labels, just that using HTML text is more bullet proof in differing scenarios that I won't list here.

@lgilders-coforma if you want to follow this approach, lmk if you'd like to pair.

## `.sr-only` class

It would be good to have a `sr-only` class in the app (really in all of our apps). They can have slight differences.

Here is the [suggestion from WebAIM](https://webaim.org/techniques/css/invisiblecontent/#offscreen):
```css
.sr-only {
  position:absolute;
  left:-10000px;
  top:auto;
  width:1px;
  height:1px;
  overflow:hidden;
}
```

Here is what [Tailwind CSS uses](https://tailwindcss.com/docs/display#screen-reader-only):
```css
  .sr-only {
    clip-path:inset(50%);
    white-space:nowrap;
    border-width:0;
    width:1px;
    height:1px;
    margin:-1px;
    padding:0;
    position:absolute;
    overflow:hidden
  }
```

I think the most important bits are breaking it out of the flow (`position: absolute;`) and moving it off screen.

services/ui-src/src/additional-scss/_globals.scss
cy-by: Your points are sound @benmartin-coforma. It's ok to lose `width: 100%; height: auto;` for these icons because those rules are meant for `` that you want to fill the container. But that's not what we want these icons to do. The only points I'd nit pick are:

1. For the footer misalignment, I'd make it more specific, something like `footer > a > .icon` or just `footer .icon` (this is pseudo code, don't quote me).
2. If we use `.large` be prepared for it to be used over places. If it's just about the PDF size, I'd name the class more specifically: `className="icon pdf"`
3. For making the PDF icon larger, I'd stay away from `transform: scale()`. Upon reading, as a CSS dev, I assume you're doing something with animations when you use `transform`. I'd put use `px` or `rem` to increase the `width`/`height`.

benmartin-coforma: 1. Rad.
2. The PDF icon shows up in different sizes in different places in the app, so we can't target by that alone. Could do a more specific class name, like `.large-icon` or `icon-3x`?
3. The `transform: scale()` is my solution to embiggen the inner SVG from the context of the outer SVG. I don't think setting the width/height on the outer SVG will do that. But maybe we could do something like `.large-icon > svg { width: ... } ` in order to target the inner SVG from the CSS? I'm just spitballing here. Lori, whatever you can make work is fine with me.

services/ui-src/src/components/EditUser/EditUser.tsx
lgilders-coforma: Replaced with new svg use.

services/ui-src/src/components/Quarterly/Quarterly.tsx
lgilders-coforma: Updated to include hidden screen reader text.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Updated to FA v5 for all icons.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Updated with span for screen reader text.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Replaced with new implementation.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Good info! Replaced with new implementation.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Good info! Replaced with new implementation.

services/ui-src/src/components/Icons/ArrowLeftIcon.tsx
lgilders-coforma: Good info! Replaced with new implementation.

services/ui-src/src/additional-scss/_globals.scss
lgilders-coforma: Updated styling with new implementation.

lgilders-coforma: The PDF Icon on the Quarterly page inherits the font size from the table stylings, so no extra classes/styling needed.

lgilders-coforma: This pre-existed my changes. I'm not sure what will happen if I remove it. 🤔  I can try it and we can see?!?

benmartin-coforma: Ah yes, it's hard to make a change like this without knowing about every button in the app. How can we know whether there's an unintended effect somewhere we're not looking?

Fortunately, SEDS is small. Here is a complete list of `<button>` elements in the app. All of them either have no icon, have a non-SVG icon, or are already touched by this PR. I believe the change is safe to make.

| Component        | Button      | Icon Type  |
|------------------|-------------|------------|
| CertificationTab | Certify (P) | -          |
| CertificationTab | Certify (F) | -          |
| CertificationTab | Uncertify   | -          |
| FormHeader       | Apply FPL   | -          |
| GenerateForms    | Generate    | -          |
| GenerateTotals   | Generate    | -          |
| Header           | My Profile  | border CSS |
| EditUser         | Update      | FA SVG     |
| FormFooter       | Save        | FA SVG     |
| FormPage         | Print       | FA SVG      |
| FormTemplates    | Save        | FA SVG     |
| Login            | EUA login   | FA SVG     |
| Login            | PW login    | FA SVG     |
| PrintPDF         | Print       | FA SVG     |
| StateSelector    | Update      | FA SVG     |
| Users            | CSV         | FA SVG     |

benmartin-coforma (commented):
I love this code, and would 1000% approve, _if_ the icons were rendering properly in [the deployed environment](https://d2kfhkd6pzpeug.cloudfront.net/). Right now it looks like this:

Image

Fortunately, it's a quick fix. The `/public` prefix on the image paths, for whatever reason (possibly some difference between the Vite dev server and the Cloudfront server?) works locally but not when deployed . Removing that prefix should still work locally, but should also work on the real site. For example:
```html
<use xlinkHref="/img/fa-icons/sign-in-alt.svg" />
```
I do apologize - I think it was my suggestion to use `/public`-prefixed paths to begin with 🤦

services/ui-src/src/additional-scss/_globals.scss
lgilders-coforma: Sadly, this breaks the styling on the login buttons that use SVGs. 😢

stacy-coforma (dismissed):
I think this is looking good 🎉 I don't see any glaring accessibility concerns, but happy to have another look from @cy-by if possible 🙂

benmartin-coforma (dismissed):
🎉

cy-by (changes requested):
## Icons not showing in some browsers

When I test in chromium, the icons are fine, but in Firefox and Safari they don't show:

Image

Image

## Non-blocking comments

More opinions, so you don't have to make changes:

* Any icon that has `aria-hidden` doesn't need `aria-label`, since it won't get read. I'd remove them so dev's don't think that `aria-label` is contributing anything to an accessible name (accName), but it doesn't hurt anything to have around.
* I [would not use "icon" in any `aria-label`](https://www.a11y-collective.com/blog/icon-usability-and-accessibility/#:~:text=For%20icons%2C%20alt%20text%20should%20succinctly%20describe%20the%20function%20of%20the%20icon%20rather%20than%20its%20appearance), you want to describe the action that icon is used to represent, not how it looks. But since all `aria-label`'s in this PR are paired with `aria-hidden` they never show, thus this is more of a nit pick.

stacy-coforma (dismissed):
checked the icons in Chrome, Firefox, Safari, and Edge just to be safe 😅

cy-by (dismissed):
Icons now showing in FF/Safari 👍🏻

benmartin-coforma (dismissed):

stacy-coforma (approved):

benmartin-coforma (approved):

